PR #9, opened 2025-03-08
Added support for Delta 3 Plus

maddios:
Used the river3.py device definition, cloned it to delta3plus.py and updated relevant info.

Works well, in my HA

image

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta3plus.py
+++ b/custom_components/ef_ble/eflib/devices/delta3plus.py
@@ -1,0 +1,123 @@
+from collections.abc import Sequence
+import logging
+from typing import Any
+
+from bleak.backends.device import BLEDevice
+from bleak.backends.scanner import AdvertisementData
+from google.protobuf.message import Message
+
+from ..commands import TimeCommands
+from ..device_properties import Field, ProtobufField, ProtobufListField, UpdatableProps
+from ..devicebase import DeviceBase
+from ..packet import Packet
+from ..pb import pr705_pb2
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class _StatField[T](ProtobufListField[T]):
+    _stat_enum = pr705_pb2.STATISTICS_OBJECT  # type: ignore[attr-defined]
+
+    @property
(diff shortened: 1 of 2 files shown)


rabits:
Hmmm, what are those unknown values in the screenshot?

maddios:
Good question, they didn't bother me but what are they even for? is that supposed to be some cumulative counter?

GnoX:
Thats integral of power calculated right on device. You would have to create these sensors yourself for the energy dashboard, but with disadvantage that it only integrates while HA is running. Delta 3 does not calculate it though.

custom_components/ef_ble/eflib/devices/delta3plus.py
```diff
@@ -0,0 +1,123 @@
...
+class Device(DeviceBase, UpdatableProps):
+    """Delta 3 Plus"""
+
+    SN_PREFIX = b"P3"
+    NAME_PREFIX = "EF-D3"
+
+    battery_level = ProtobufField[int]("bms_batt_soc")
+
+    ac_input_power = ProtobufField[float]("pow_get_ac_in")
+    ac_input_energy = _StatField[int]("STATISTICS_OBJECT_AC_IN_ENERGY")
```
rabits, line 48: So yeah, if the device can't calculate those fields - you need to get rid of them or prepare a logic to calculate it within the integration, although that could be a bit complicated...

rabits (changes requested):
Sorry, we can't have the device with some missing fields - but I'm sure you can just remove the unused fields to make it minimal to not confuse the other users.

Anyway thank you for preparing the change and testing it! It means a lot for the owners of Delta 3 Plus.

MansM:
@Maddies can't wait to see it being integrated.

rabits:
Oh, sorry @maddios - you need to check https://github.com/rabits/ha-ef-ble/pull/12 - it's already merged as [0.4.2](https://github.com/rabits/ha-ef-ble/releases/tag/v0.4.2) and this one was abandoned, so I will just close it.

Issue #5, opened 2025-02-21
Add suppart for EcoFlow River 3 devices


BenBergman:
I have two River 3 units I'd love to connect to HA through BLE. Both units are still on factory firmware 1.0.0.0 with 6.20.48.68 listed as available.

I have a Linux laptop and can work with the reversing repo to help add device support if there is interest in adding it.

rabits:
Hi @BenBergman ! Thank you for the offer - for sure, if you can create logic to connect River 3 and support it - I will be able to help you to integrate it into the existing project structure. Hopefully it will fit for this device, but if not we will find a way.

rabits:
Regarding the firmware - I was unable to start from 1.0.0.0 for my devices, so updated it to the latest available on the date. But if you want to try to support 1.0.0.0 and 6.20.48.68 - that's a good start to have just 1.0.0.0, but it probably it will not support the ecoflow BLEv2 protocol in this fw...

BenBergman:
I don't have any specific reason to prefer firmware 1.0.0.0, I just figured it might be helpful to have an old version since sometimes companies start to lock things down. I'll update one of them to the latest firmware and work from there and then test it on 1.0.0.0 on the other. If it works, great! If not, no big deal.

I'm trying to use the reversing tool now but I'm not totally sure what I'm looking at. I had to tweak the script to decode the serial before checking if was supported, otherwise it was crashing.

```
diff --git i/connect.py w/connect.py
index d0a45d9..69e2cf5 100755
--- i/connect.py
+++ w/connect.py
@@ -95,8 +95,8 @@ class Device:
         # Looking for device SN
         man_data = adv_data.manufacturer_data[Device.MANUFACTURER_KEY]
         sn = man_data[1:17]
+        self._sn = sn.decode('ASCII')
         if sn[0:4] in Device.SUPPORTED_DEVICES:
-            self._sn = sn.decode('ASCII')
             print("%s: Parsed SN: %s" % (self._address, self._sn))
         else:
             print("%s: Unknown SN: %s" % (self._address, sn))
```

Now it says I'm connected, but I keep getting warnings about data packets not being processed.

```
9C:9E:6E:59:9E:16: Common: Device connected & ready: Packet(0x35, 0x21, 0x01, 0x52, bytes.fromhex('ee00ba67fa'), 0x01, 0x01, 0x03, b'\x00\x00\x00\x00', 0x00)
[...]

PR #10, opened 2025-03-08
Add controls for River 3/Plus, add SN and mac to device info

GnoX:
This PR adds switches, numbers and selects for River 3, changes name to correct variant based on serial number and also adds more information into device info.

### DIFF
--- a/custom_components/ef_ble/eflib/devices/river3.py
+++ b/custom_components/ef_ble/eflib/devices/river3.py
@@ -45,7 +45,7 @@ def _flow_is_on(x):
 class Device(DeviceBase, UpdatableProps):
     """River 3"""
 
-    SN_PREFIX = b"R6"
+    SN_PREFIX = (b"R651", b"R653", b"R654", b"R655")
     NAME_PREFIX = "EF-R3"
 
     battery_level = ProtobufField[int]("bms_batt_soc")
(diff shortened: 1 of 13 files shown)


rabits:
Nice change, I will check it closer when it will be ready for review.

GnoX:
> Nice change, I will check it closer when it will be ready for review.

I'm Just waiting for ack from any Plus owner, otherwise its ready.

GnoX:
Got confirmation in https://github.com/rabits/ha-ef-ble/issues/5#issuecomment-2711487179, PR is ready for review.

custom_components/ef_ble/eflib/devices/river3.py
```diff
@@ -88,6 +122,7 @@ async def data_parse(self, packet: Packet):
             p = pr705_pb2.DisplayPropertyUpload()
             p.ParseFromString(packet.payload)
             _LOGGER.debug("%s: %s: Parsed data: %r", self.address, self.name, packet)
+            _LOGGER.debug("River 3 Parsed Message \n %s", str(p))
```
rabits: Some additional debugging? Do we need it or maybe to comment/remove?

rabits (changes requested):
Very nice change, just probably you forgot additional debugging there.

custom_components/ef_ble/eflib/devices/river3.py
GnoX: I actually included this intentionally for folks who want to help with identification of fields so that they don't have to run it through connect script, but I realized that with multiple devices, it will completely flood the logs. Better way to do it would be to add a checkbox to reconfigure flow that enables/disables logging for individual devices. 

Anyways, thanks for catching this, I will remove it for now.

GnoX: Removed in [8204756](https://github.com/rabits/ha-ef-ble/pull/10/commits/8204756e887e93cdd04a55b67e7e9474211d817b)

rabits: Thank you! Yeah, better will be to have some option for that.

PR #11, opened 2025-03-08
Rework config flow

GnoX:
This PR fixes config flow strings, adds fetching user id automatically based on ecoflow login info and adds validation for user id (connects to device before adding entity) .

Filling in login details and pressing submit will set user id. Submitting without login info set will validate user id by connecting to the device and sets error message if something fails. After user id is validated, it is saved to disk cache and will be reused for all subsequent devices.
Screenshot from 2025-03-08 22-12-24


rabits:
Hmmm, interesting change... Let me think about it a bit.

rabits (approved):
In general - no objections, just worried about "what if user will rely on the email/password to connect it's device to the remote service and if service dies in 10y he will never figure out hiss user id". But I think it's much better approach then pointing users to the documentation on "how to get your user ID through developer console".

I appreciate you making this poor integration look much better, thank you very much)

rabits:
Let me quickly test this change tomorrow and I will merge it.

GnoX:
> just worried about "what if user will rely on the email/password to connect it's device to the remote service and if service dies in 10y he will never figure out hiss user id"

That's a valid concern, but we can very easily disable it if it comes to it :)

> But I think it's much better approach then pointing users to the documentation on "how to get your user ID through developer console".

Yeah, even with the documentation, people still have some problems finding the user id so this will hopefully make your job a bit easier.

custom_components/ef_ble/config_flow.py
```diff
@@ -60,22 +71,43 @@ async def async_step_bluetooth_confirm(
...
                 {
-                    vol.Required(CONF_USER_ID): str,
-                    vol.Required(CONF_ADDRESS, default=device.address): str,
+                    vol.Optional(CONF_USER_ID, default=self._user_id): str,
+                    vol.Required("login"): section(
+                        vol.Schema(
+                            {
+                                vol.Optional(CONF_EMAIL, default=self._email): str,
+                                vol.Optional(
+                                    CONF_PASSWORD, default=self._password
```
rabits: Hey, just found you keeping email/password in memory to reuse it for the other devices - maybe instead to keep the user id and wipe-out email/password from memory? Probably we don't need to keep the email/pass when we detected the user id?

rabits (changes requested):
Took closer look and my security tooth started to bug me. Don't really like we keeping login/password in memory since we need it just one time to figure out the user id.

custom_components/ef_ble/config_flow.py
GnoX: Method  `_ecoflow_login` always sets email and password to empty string upon successful auth. Also, `_ecoflow_login` is always called if you filled either the username or the password so there should be no way to sucessfully pass through config flow with non-empty username and password. 
Or am I overlooking something?

rabits: So you see you use those vars as defaults, that makes me think "why" if those are empty. Maybe instead of keeping them in self._email / self._password you can do the same with self._user_id and keep it for the other devices and just store email and password in local vars to make sure they will not be saved for long?

GnoX: I implemented it like this because I found it quite irritating to have to reenter credentials on each error, but it probably makes sense to always clear the password.

BTW most integrations store credentials and tokens in plain text anyways, afaik HA does not have a way to store them securely.

GnoX: > So you see you use those vars as defaults, that makes me think "why" if those are empty. 

I did not find another way to save them after you click the submit button and if error happens, you have to fill everything again.  I guess it is not a big problem though.

> Maybe instead of keeping them in self._email / self._password you can do the same with self._user_id and keep it for the other devices and just store email and password in local vars to make sure they will not be saved for long?

I mean, I get your point, but if I save them as local vars, they are still in memory until they are garbage collected so they may actually stay there longer than in current implementation where I set them to empty strings.

Anyways, I'll change it so that password is always cleared.

GnoX: I changed so that password is only stored in local variables. I don't think it matters much though, it will still be in the memory until it is garbage collected. Should I do it properly and set it to empty string in the `user_input` and `password` or is this sufficient for you?

https://github.com/rabits/ha-ef-ble/pull/11/commits/550e1af943bb0386039f619fddc9c9c327d0c5e6

rabits: Yeah I see your point now, but still looks like too much for just process the user wrong input and show the user what he put there... So I think if we will keep it simpler with no complicated logic - it will be much easier for the newcomers to understand of what's happening in the logic.

Another option could probably be commenting on those tricks you using to make it easier for reader to understand that those are not stored for long here, but since you removed this storage - I think it's not needed now) Thank you!

rabits:
Works very nice! I really like how it's relaunching the window with entered user_id. Tested on my SHP2/DPU devices - works perfect)

rabits (approved):
Tested & works great

GnoX:
Great! Thanks for review.

PR #12, opened 2025-03-15
Add support for Delta3, improve typing for protobuf fields

GnoX:
This PR adds support for Delta 3 (tested of Delta 3 Plus) and does some refactoring of fields that makes them tied to typing stubs and reduces chance of misspelling or using the wrong type and makes them easier to find and define.

It also fixes generated protobufs for River 3 - previous version does not have a namespace which made collisions with Delta 3 protocol buffers.

Screenshot from 2025-03-16 17-58-03


custom_components/ef_ble/eflib/pb/pd335_bms_bp_pb2.pyi
```diff
@@ -0,0 +1,614 @@
+from google.protobuf.internal import containers as _containers
```
rabits, line 1: Hmmm do we need those pyi files here?

custom_components/ef_ble/icons.json
```diff
@@ -0,0 +1,13 @@
...
+            "battery_sync": {
+                "default": "mdi:battery-sync",
+                "state": {
+                    "off": "mdi:battery-sync-outline"
+                }
+            }
+        }
+    }
+    
+}
```
rabits: POSIX text lines requires newline char in the end

rabits (changes requested):
Nice change, just needs some small adjustments

custom_components/ef_ble/eflib/pb/pd335_bms_bp_pb2.pyi
GnoX: Yeah, I kinda expected this question considering how long they are. These are generally paired and tracked with generated protobufs so I included them as they make development with protobuf messages easier as they now autocomplete and display types.

Unless you have a strong opinion against it, I would keep them in.

custom_components/ef_ble/icons.json
GnoX: This just shows how reliant I am on pre-commit checks in my projects :)

Fixed in [65c2177](https://github.com/rabits/ha-ef-ble/pull/12commits/65c2177f527d2706aea7eb94c4788251d5ae8e6d)

custom_components/ef_ble/eflib/pb/pd335_bms_bp_pb2.pyi
rabits: Hmmm, ok no problem, let's keep them)

custom_components/ef_ble/icons.json
rabits: Np)

rabits (approved):
Ok looks good, tell me when it's ready for merge

GnoX:
Ready from my side, thanks for the review.

PR #13, opened 2025-03-22
Add configurable device update period to reduce cpu usage and writes

GnoX:
As per [this request](https://community.home-assistant.io/t/ecoflow-ble-unofficial/774794/88?u=gnox), this PR adds option to specify update period to devices - only one packet is parsed per period. This reduces cpu usage for devices that receive a lot of images (i.e. Delta 3 receives 3 messages per second).

I also removed typing stubs from the previous PR. What I didn't realize is that they unnecessarily take 250KB in HA. It would be better if we added proto definition files here and provided script or guide on how to generate them for dev purposes.


custom_components/ef_ble/eflib/connection.py
```diff
@@ -516,17 +531,49 @@ async def autoAuthentication(self):
...
-        payload = ("".join("{:02X}".format(c) for c in md5_data)).encode("ASCII")
+        payload = ("".join(f"{c:02X}" for c in md5_data)).encode("ASCII")
 
         # Forming packet
         packet = Packet(0x21, 0x35, 0x35, 0x86, payload, 0x01, 0x01, 0x03)
 
         # Sending request and starting the common listener
         await self.sendPacket(packet, self.listenForDataHandler)
 
+    def _can_update(self):
```
rabits, line 588: Hmmmm, this one is bothering me... SHP2 uses multiple different protobufs (https://github.com/rabits/ha-ef-ble/blob/main/custom_components/ef_ble/eflib/devices/shp2.py#L95-L165) - so could you please tell how it will work here? For example we can receive 1(2 or 3 ) message(s) about the circuits and one about battery level...

rabits (commented):
Oh those pyi files... But if they are not needed - yeah better to remove. And this is surely needed change to reduce the workload, but 3 questions:
1. Will it really help? How it affects CPU utilization on your end? On my RPi 4 with 2 SHP2 devices (and multiple others) it takes ~10-15% of CPU (loadavg 0.80, docker stats shows 30-40%) - so I will probably be able to tell if that helps...
2. I suppose parsing of the message is a pain, but could we really get away with such simple counter implementation? I did not search for notification throttling, but maybe the protocol supports it?
3. Worrying the other devices would not like that, but I will try to help with them - don't want to have some configs working for some devices and some don't.

GnoX:
> Will it really help? How it affects CPU utilization on your end? On my RPi 4 with 2 SHP2 devices (and multiple others) it takes ~10-15% of CPU (loadavg 0.80, docker stats shows 30-40%) - so I will probably be able to tell if that helps...

I did run a profiler and found that most of the time was spent in packet decrypting and field assignments so using higher update time should have reduced the usage. I did a clean HA install on my rpi4 and it does, in fact, reduce cpu usage, but only very slightly. I wonder why was the usage so high for the guy that requested this - rpi4 cpu should be worse than what he has. 
Maybe we can only present this option in reconfigure and not in the main setup and see if this change helps?

> I suppose parsing of the message is a pain, but could we really get away with such simple counter implementation? I did not search for notification throttling, but maybe the protocol supports it?

So, this is the first time I code something for bluetooth and not sure what is the correct way to handle things. I did find any way to throttle incoming messages provided by the bt library and from what I read, the sending rate is only controlled by the device. I tried sending [some config options](https://gist.github.com/GnoX/4e9631ba8f9f1942aeb2774dd0a93662#file-pr705-proto-L455) to control upload time, but I guess these only apply to wifi upload. If you have any other suggestion, I'll be happy to try it.

> Worrying the other devices would not like that, but I will try to help with them - don't want to have some configs working for some devices and some don't.

I can probably only enable this config option for devices I own as this was only requested for River 3 at the moment.

custom_components/ef_ble/eflib/connection.py
GnoX: Hmm, are they sent at different intervals? Yeah, this is not going to work then. D3+ and R3 do always send 3 and 2 messages respectively. What we could probably do in this case is throttle packets based on the command labels but this is a bit questionable as decryption can't be skipped then.

GnoX: After thinking about this a bit more, I can probably try to do it like this:
1. instead of throwing away the packets, store them encrypted in a buffer
2. after throttle period, pass a generator that decrypts packets as they are being iterated over in reverse and device can decide to stop iterating and do an update
3. clear the buffer, repeat from step 1

Do you think something like this would be ok?

rabits: Hmmm, I wonder if we can actually run decryption on the combination of packets at the same time... But probably not, so imagine battery update comes every 10 seconds and the rest of updates are coming every 0.3 seconds... So without decryption that will be hard to catch. Maybe we can try to utilize the available hw decryption?

rabits:
@GnoX I think you can try to ask for the initial measurements from the requestor - so we will have a proper baseline to compare with. Maybe he mentioned just one vCPU he has, like on mine it shows docker stats 30-50% utilization of one vCPU and it will show 400% if all 4 vCPUs will be used?

Yeah, probably there is not much we can do to throttle the message rate on the device, but let me look into - maybe I will find something helpful...

Maybe ask him to test the change first before merging - so at least we will see it's actually helping? Because for now I see no easy way to solve this issue with encrypted packets...

custom_components/ef_ble/eflib/connection.py
GnoX: So, I did change the implementation to accumulate encrypted packets for specified time period and decrypt them on demand controlled by device implementation. For the example you mentioned, it won't have a major effect as you have to decrypt most of the packets, but can still choose to not parse them. However for R3 and D3, there's always 3 messages in sequence so with this approach, we only need to decrypt and parse last 3 packets and ignore the rest. 

It also should not break `shp2` and `dpu`  but I'm not sure what happens if you set `update_period` for them - I think it will work and write only the last value, but callback is going to be called for each packet.

rabits: Yeah, so did you had a chance to ask the requestor to test the change - does it help him?

GnoX: I didn't yet, but tried to get measurable usage in a docker container constrained to 0.5 cpus. I saw reduced load by relative ~35% with 5 second update period. And as expected, with higher update period, it is even higher. I don't believe though that on a modern cpu, even single core, cpu usage can get too high by installing this addon (unless you have a lot of ecoflow devices). What I'm getting is that if this won't help the reqestor, I don't think anything else we can do will :)

That said, we should still have at least throttle writing to db - I noticed that it grows quite quickly with the default 1 second writes. I could just revert this and implement simpler write throttling. I'm already using current implementation on my main HA machine and appears to be working fine (at least with 3 second update period), but I don't mind reverting it. Ultimately, I'll leave that decision to you.

rabits: Ok, let me try to contact him to test this approach.

PelleRavn: Would love to test it and give some measurements. But I'm not that familiar with HACS, so what's the best way to test this in HA?
I can add as a baseline, my VM (with 2 cores at 1,5 GHz) it uses around 32% without the project running. After it goes to 37-41%. CPU load over 1 minut went from ~5,5 to ~6.4.

GnoX: @PelleRavn I have released these changes as a prerelease [v0.10.0-a.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.10.0-a.1) in my fork, so you can just swap hacs repo temporarily to test it.

PelleRavn: @GnoX Seems to be working great. Change it to update every 5 (seconds I guess). When setting it to 5, the average CPU load goes a bit down, roughly to 35%. Think that's better, but honestly, no matter the CPU usage, I think it's worth lowering it - so great work!

rabits: @PelleRavn you mean the average (for 1min for example) or instant load goes down to 35% from 65%? If it's the average - then it's not "a bit", but "quite a bit", because your original was 20-30% load, right?)

PelleRavn: > @PelleRavn you mean the average (for 1min for example) or instant load goes down to 35% from 65%? If it's the average - then it's not "a bit", but "quite a bit", because your original was 20-30% load, right?)

No I think I expressed it wrong. 
The thing is, it's a bit hard to measure since HA doesn't have a way to show what it is using. So I'm reporting what my Hypervisor says the VM is using.
And it went from ~32% when it wasn't running, till 37%-41% when it is running, so 5-9%. But again, that's pretty relative.
But I have had 2 issues with HA starting to fail recently. All integrations failed to load and nothing is working (other than restarting it). I'm unsure if all the processing is making the system go out of memory and then everything start getting killed by the OS. So now I have disabled the integration to see if it happens again without it running. Happed with 6 days between the events, so will have to see.

rabits: Thank you @PelleRavn for testing, yeah we understand it takes time. Maybe there is a buffer that collects too much data, so better to be safe.

PelleRavn: @rabits After I disabled the integration, HA have been running fine. 
So I think there is something that is leaking causing HA to run out of memory. Any ideas on how I can test and verify that?

GnoX: Does it happen only with version I posted or does it happen with current release version? I may have introduced memory leak with these changes. Profiling is not very easy when running in HA as I didn't find an easy way to limit it to a single integration, so I'll probably try profiling it myself in isolated HA instance.

GnoX: Ok, I ran some profiling and identified that there may be a leak, but it seems to be very slow. I identified something more concerning though, looks like we should not run `async_write_update_state` for each state update callback - most of the allocations originated from there. What we should do is use [DataUpdateCoordinator](https://developers.home-assistant.io/docs/integration_fetching_data) (or ideally [bluetooth coordinator](https://developers.home-assistant.io/docs/core/bluetooth/bluetooth_fetching_data)) that only calls update once for all fields.

Also, here's how the memory size looks when the integration is enabled vs disabled.
|  Disabled | Enabled |
| ------------------ | -------------- |
| Screen Shot 2025-05-04 at 16 15 01 | Screen Shot 2025-05-04 at 16 16 06 |

rabits: Great find @GnoX ! I actually played with those coordinators before, they could work in theory - just one issue that they are trying to reconnect for each round and then drop the connection. I got around that in my another integration, but not quite sure it's worth the hassle yet...
What I see on the graph is that it reaches the threshold of GC and collected. That's for sure takes more cycles for GC, interesting what will happen if the GC threshold will be increased - where it will raise... And overall for sure it's understandable to fluctuate for 10-20MB to decrypt the incoming packets... Quite curious to see - how this change will fix the graph?

GnoX: > just one issue that they are trying to reconnect for each round and then drop the connection

Is that the case for all types of bluetooth coordinators? Looking at the documentation again, I guess we don't receive advertisement messages but only active bluetooth messages so none of them are applicable, only custom DataUpdateCoordinator (I see that this is what you did this in your other integration).

> Quite curious to see - how this change will fix the graph?

I think I had the wrong understanding of how coordinators work, I thought that they also pool HA state updates, but that does not seem to be the case, we still have to call the write state, so I guess using coordinator would not really help.

I had this idea while I was looking at a temporal segment of memory allocations filtered to only show code from this integration - we can see that the actual decryption only allocates 7.2KB over the span of 30 seconds while call to `async_write_ha_state` allocates over 100KB with 700 allocations over the same period. 
Screenshot from 2025-05-05 00-34-02

I'll see what we can do about it, I may be doing something wrong.

GnoX: > just one issue that they are trying to reconnect for each round and then drop the connection

Is that the case for all types of bluetooth coordinators? Looking at the documentation again, I guess we don't receive advertisement messages but only active bluetooth messages so none of them are applicable, only custom DataUpdateCoordinator (I see that this is what you did this in your other integration).

> Quite curious to see - how this change will fix the graph?

I think I had the wrong understanding of how coordinators work, I thought that they also pool HA state updates, but that does not seem to be the case, we still have to call the write state, so I guess using coordinator would not really help.

I had this idea while I was looking at a temporal segment of memory allocations filtered to only show code from this integration - we can see that the actual decryption only allocates 7.2KB over the span of 30 seconds while call to `async_write_ha_state` allocates over 100KB with 700 allocations over the same period. 


I'll see what we can do about it, I may be doing something wrong.

rabits:
Actually I thought more about that in the light of #28 - we can actually not send replies and this will make SHP2 (and probably other devices) to not send additional data back - since we replying there with every processed message - maybe we can play with that to allow users to reduce encoding/decoding with this option. I'm actually not sure which replies have to be to get the data - but can play with that and see what kind of updates are coming.

Or maybe even manage the amount of replies we sending? So that potentially should keep everyone happy - still get frequent updates on important stuff and receive the less important updates when needed - because as far I know if you stop to send replies - it stops to send additional data as well.

GnoX:
That may work for SHP2 and DPU, but not for D3 and R3 - if you look at the implementation, these are not sending back replies at all. 

What I'll do to speed this up is to check if the memory profile is the same without this change - if it is, we can probably merge this and solve memory issues later. If it is not, what I'll do is to focus on storage size rather than cpu usage and implement period only for state callbacks, not decryption.

GnoX:
I looked at the memory profile without these changes and it seems that peaks are a bit higher with deferred decryption, but garbage collection was slightly more spread out suggesting that it did help albeit not significantly. Memory leak is a bit concerning, but I am not able to find it easily, it must be very slow. I'll have to run hass under profiling for much longer time to find it.

For now, I'm putting this back into draft - #30 should at least fix DB balooning problem, but CPU is a bit harder to test as I am getting a bit incosistent readings on all of my devices as the cpu usage is very low. I will revisit this later if we get more reports, but it may be too much effort for potentially isolated case.

rabits:
> That may work for SHP2 and DPU, but not for D3 and R3 - if you look at the implementation, these are not sending back replies at all.

Yeah you right... As usual I think about that from my tower) If you want I can share with you the modified apk/jadx dict so you will be able to see the logs and capture ble via android tools - so maybe some new idea that I've missed will come to you on how to properly throttle that process... And I will try to find something - I don't believe they don't have something to control the flow.

GnoX:
> If you want I can share with you the modified apk/jadx dict so you will be able to see the logs and capture ble via android tools - so maybe some new idea that I've missed will come to you on how to properly throttle that process... 

I've already looked through a good amount of deobfuscated java code related to bluetooth and found nothing that would be relevant. It is probably possible through some hidden command, but there's nothing in the app that could control it. At least I didn't find anything useful (I didn't look at the native code though, only java).

PR #24, opened 2025-04-21
Add AC charging speed controls for D3, R3, DP3 and DC controls for D3, R3

### LINKED ISSUE #23: Allow AC Charging Speed to be set
Hi all,

First off a big thank you for this integration, I've been using it with my new River 3 Plus to load shift solar power and it's working really well!

I noticed in the app under Input and output settings there is an AC charging speed screen where I can set the charging speed from 80w to 640w (380w without the second expansion battery) in 1w increments. Is it possible to allow setting this option as well?

The one issue I have with my automations is quite often my solar is providing enough power for the house in the morning's or on overcast days but isn't exporting enough to begin charging the battery (640w + device load). Is it possible for the integration to allow setting the charging speed so the battery can be charged at slower speeds? The other option I have is to manually set a lower charging speed but this wouldn't be able to make as much use of spikes in solar output.

### LINKED ISSUE #25: Set DC charging speed (car input)
On my River 3, I can set DC charging mode to Car or Solar. When set to "car recharging", the input can be set to 4A, 6A or 8A.

I love the option to set AC charging speed (#23) but would like to set the DC speed as well. It would probably make sense to also have the DC charging mode configurable. 
My personal use case: the DC input is connected to an LFP battery (which is charged separately). I want to reduce stress on both batteries by transferring at the lowest rate. Only when the power use demand is higher than 4A, I want to increase the charge rate to 6A.

(I'm currently waiting for a Zigbee DC switch so I can even disconnect the DC input, so I can prevent the River 3 from charging over 25% by DC if I want to import cheap AC power later in the day)

Image

GnoX:
This PR implements controls for setting AC charging speed as requested in #23 for River 3 (tested), Delta 3 (tested on D3+) and Delta Pro 3. The Ecoflow app does limit input values for some devices and sets lower limits, but it does still work (at least on R3 and D3+).

It also includes DC charging controls for D3 and R3 requested in #25

Resolves #23
Resolves #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,10 @@ Recognized devices:
 | *Sensors*                       | *Switches*     | *Sliders*            | *Selects*            |
 |---------------------------------|----------------|----------------------|----------------------|
 | AC Input Energy                 | AC Port        | Backup Reserve Level | Led Mode (Plus only) |
-| AC Input Power                  | DC Port        | Max Charge Limit     |                      |
+| AC Input Power                  | DC Port        | Max Charge Limit     | DC Charging Type     |
 | AC Output Energy                | Backup Reserve | Min Discharge Limit  |                      |
 | AC Output Power                 |                | AC Charging Speed    |                      |
-| Main Battery Level (Plus only)  |                |                      |                      |
+| Main Battery Level (Plus only)  |                | DC Charging Max Amps |                      |
 | Battery Level                   |                |                      |                      |
 | DC 12V Port Output Energy       |                |                      |                      |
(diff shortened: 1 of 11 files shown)


rabits:
Hi @GnoX , the changes looks perfect as usual) Please call me to review when all the changes will be here - just worrying you still working on it...

GnoX:
Hey, you can safely assume that once I take it out of draft it is not only ready to be reviewed but also merged.

rabits (approved):
As usual looks very nice!

PR #30, opened 2025-05-10
Add DB write period configuration option to all devices

GnoX:
This PR is simplified code from #13 that does not modify packet processing, it modifies mesage processing in devices instead. We still decrypt every message, but only writes new state to DB after configured update period and only deserializes message if it needs to.

The logic for updating fields is little bit more complicated as it has to handle the same message type but with different attributes. This is solved by specifying discriminator fields that determine message types. This can however result in updating sooner than the update period if fields are not disjunctive between the same message type, but it won't increase number of writes from previous version.


GnoX:
Actually, just as I finished this I got a much simpler idea and I'm mad at myself that I didn't notice this a bit earlier at it seems like an obvious solution now - we can just store all modified fields within period and run callbacks only once. While this won't solve cpu usage, it definitely solves db usage.

rabits:
Yeah will check that today, thank you very much! I also thought about changing this update rate with options - something that will not require reloading of all the integration devices - wdyt? For example when we need detailed info - just changing option and it's applied automatically... I saw it in some integration, but forgot where...

rabits (approved):
Overall the change looks fine and I tested it, works smoothly - previously I tried to make some sort of interpolation mechanism to check if the points are forming a straight line - then skip the new point until it deviates by 20-30% and only then write to db. This way it should still show the immediate changes, but not write too often and parameter could be in what diapason to tolerate the deviation. But unfortunately 2 problems: computational and I'm not that good in math) So maybe you will be up to the challenge, just an idea - but for now it's perfect)

GnoX:
> I also thought about changing this update rate with options - something that will not require reloading of all the integration devices - wdyt?

Ohh, I remember deleting options flow code in here as it was not used here and not gonna lie, I didn't see how it differed from config flow just from reading the [docs](https://developers.home-assistant.io/docs/config_entries_options_flow_handler/). It does make more sense to have it as an option flow entry, though it would be nice to be able to change it from initial config screen as well. This is possible, but makes handling more complicated, so I'll skip it for now and implement it just as an option.

> previously I tried to make some sort of interpolation mechanism to check if the points are forming a straight line - then skip the new point until it deviates by 20-30% and only then write to db. This way it should still show the immediate changes, but not write too often and parameter could be in what diapason to tolerate the deviation. 

I tend to overcomplicate things and I did in fact want to implement something like per property throttle where it would wait for the closest message if it comes in few seconds to always make it be very accurate, but opted for just writing the latest value at the time of update.

I may look at some statistical approach in the future, maybe something similar to what you tried, but I think the current method will suffice for now.

GnoX:
It was actually not that complicated to show it in both initial config and options so it is now in both. Ready to merge from my side.

rabits:
Just checked, exactly how I thought about it - configure button and no device reload. You're wizard man, thank you very much!)

PR #29, opened 2025-05-08
Use fields for DPU and SHP2, add individual battery sensors and AC port sensors to DPU, use translations for all sensor icons and names

GnoX:
This PR adds a way to use multiple messages types for pb_fields and uses pb_field defs from latest devices for DPU and SHP2 making it quite easy to add new fields. It also moves all sensor names and icons to translation jsons.

PR #31, opened 2025-05-14
Fix DPU not updating after update interval change

GnoX:
In #29 I forgot to update updated variable for updated properties - this caused DPU to not update in GUI.


rabits (approved):
Hmmm, weirdly but my DPU works just fine... But ok, since this should help the others)

rabits:
Oh wait, then you will not need updated_props on https://github.com/rabits/ha-ef-ble/blob/main/custom_components/ef_ble/eflib/devices/dpu.py#L78 ?

GnoX:
That's the screw up I did - before I updated this to the field style, it was appended into that local list, but with field style  names are appended into `self.updated_fields`.

PR #32, opened 2025-05-18
Add options to control logging output

GnoX:
This PR adds options to enable debug logging for specific devices and groups messages into categories so only relevant info is displayed when debugging. It also adds option to mask mac addresses, serial numbers and user ids for pasting these into issues or forums, you never know who's watching.

### DIFF
--- a/hacs.json
+++ b/hacs.json
@@ -1,5 +1,5 @@
 {
   "name": "EcoFlow BLE",
   "render_readme": true,
-  "homeassistant": "2024.6.4"
+  "homeassistant": "2024.7.0"
 }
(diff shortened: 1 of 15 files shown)


hacs.json
```diff
@@ -1,5 +1,5 @@
 {
   "name": "EcoFlow BLE",
   "render_readme": true,
-  "homeassistant": "2024.6.4"
+  "homeassistant": "2024.7.0"
```
GnoX, line 4: For some reason, HACS validation job started failing so I tried bumping version and it works now, no idea why.

rabits: Ok, little bump works and that's fine)

rabits (approved):
Very nice! Thank you, that surely will help folks to find out "why it's not working"

PR #33, opened 2025-05-21
Fix SHP2 grid power not updating on 0, clean up logging

GnoX:
This PR introduces these changes
- SHP2 grid power sensor should now be updating when it changes to 0 -This was fixed by introducing `TransformIfMissing` class that can be provided to transform function of `pb_field` and this tells the field to process the value even if it is not present in protobuf message - in this case it will receive None as the value to transform
- cleans up individual device logging that was still using global logger instead of device loggers
- adds masking filters to root logger instead of local logger as other loggers may also print mac addresses and serial numbers
- removes dev only pyi files


rabits:
Yep, will check tomorrow, thank you)

rabits (changes requested):
Those pyi files getting through anyway... So I see you removing 3 of them, but we have another 3 here:
* custom_components/ef_ble/eflib/pb/pd335_bms_bp_pb2.pyi
* custom_components/ef_ble/eflib/pb/pd335_sys_pb2.pyi
* custom_components/ef_ble/eflib/pb/pr705_pb2.pyi

Do you think it's time to remove them and add to gitignore? Since it's not going to be fixed in the dev env - I'm now leaning towards adding them to project's gitignore...

But otherwise it works fine)

GnoX:
I'm not sure what you mean - are you looking at the correct branch? All 3 stub files you listed should have been removed in [this branch](https://github.com/GnoX/ha-ef-ble/tree/fix-shp2-grid-power-zeroing-and-clean-up-logging/custom_components/ef_ble/eflib/pb).

rabits:
Yeah, sorry, you right - I compared with main) I should stop to review changes during the night time)

rabits (approved):
LGTM

Issue #42, opened 2025-06-07
[Bug]: CPU usage gradually increases


chalimov:
### What happened?

The system's CPU usage gradually increases over time, rising from approximately 2% to 10% over the course of 2 to 3 days, after which I typically restart the system.

### What version of our software are you running?

0.4.13

### What device are you seeing the problem on?

I have two Delta 3 and one Delta 3 PRO systems.

### What bluetooth connection are you using?

ESPHome bluetooth proxy

### Integration settings (if you changed them)

```text

```

### Relevant log output

```shell

```

GnoX:
Hello,

this may prove to be a bit annoying to diagnose as it does not seem to happen on my HA instance that is up for 34 days with 2 devices. I'm currently not sure what would be causing the rise though, I'll try to do some profiling.

Also, we are aware of the higher CPU usage for some cpus, but it is a bit difficult to solve as we cannot control how many messages are sent from the device - for example River 3 sends 3 messages per second, each has to be decrypted before we can check if the value changed. There is a way to reduce it a little bit for devices with non-incremental message updates (R3,D3,D3P) - wip #13. Other than that, unless we figure out how to tell the device to reduce the number of messages, I don't think there's anything else we can do.

chalimov:
Hi,

Thank you for your response. Just to clarify, the issue I'm experiencing is not a high(er) CPU usage. but a steady increase of it. When I disable the devices, the CPU usage goes down to 1% and stays there. If there is anything I can do on my side to help you  diagnose the issue, please let me know.  And thank you for doing this!

GnoX:
Oh, sorry, that "higher CPU usage" paragraph was more of a sidenote to explain what we're dealing with :). I did understand what your issue is.

What we'll probably need is few profiler outputs when the CPU usage is lower and when its higher from [profiler](https://www.home-assistant.io/integrations/profiler/#action-profilerstart) integration collected over the same amount of time - probably around 5 minutes should do. With several of these samples, we should be able to at least check what functions are taking longer and hopefully identify the cause. 

I do have few ideas what it could be, but can't be sure as I'm not able to measurably reproduce it with my devices.

BenBergman:
I believe I'm having the same issue. I have 2 River 3 units, one directly next to my HASS server with built-in bluetooth and the other at the opposite end of the house next to an ESPhome device with bluetooth proxy.

I've got a bunch of these lines in my HASS Core logs:

```
[...]

Issue #36, opened 2025-05-25
[Bug]: Error Logging Loop / Disk Space Usage


some-guy-23:
### What happened?

I installed this integration yesterday, and woke up today to find a huge increase in disk space usage in Home Assistant, also just seemed to bog down my entire system with lots of connection issues on unrelated integrations. Traced it to this integration, specifically logged something ~3.5 million entries of the error seen at bottom.

Only way I was able to reclaim the disk space was to first delete home-assistant.log.1 (after reboot the old log is moved here). Then had to also copy out the HA database and run the following script on it and then copy it back (commented out actual deletes for dry-run first) -

`CREATE TEMP TABLE to_delete_data_ids AS
SELECT events.data_id
FROM events
JOIN event_data ON events.data_id = event_data.data_id
WHERE events.event_type_id = 48
  AND event_data.shared_data LIKE '%custom_components/ef_ble%';

SELECT count(*) FROM to_delete_data_ids;

-- DELETE FROM event_data
-- WHERE data_id IN (SELECT data_id FROM to_delete_data_ids);

-- DELETE FROM events
-- WHERE data_id IN (SELECT data_id FROM to_delete_data_ids);

DROP TABLE to_delete_data_ids;

-- VACUUM;`

### What version of our software are you running?

0.4.10

### What device are you seeing the problem on?

River 3 Plus

### What bluetooth connection are you using?

ESPHome bluetooth proxy

### Integration settings (if you changed them)

```text

```

### Relevant log output

```shell
[...]

PR #46, opened 2025-06-19
Fix error spam due to forgotten error counter

GnoX:
I may have accidentally removed error counter incrementation somewhere in my previous PR's causing logs spams, this may have been the cause of both #42 and #36.


rabits (approved):
Aiaiai, but it's great you found it! Thank you)

PR #49, opened 2025-06-26
Remove Delta 3 1500 from supported devices

GnoX:
I incorrectly assumed that Delta 3 1500 uses the same protocol as other Delta 3s but it seems that it uses Delta 2 firmware which probably makes sense as the port configuration is identical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Recognized devices:
 </details>
 
 <details><summary>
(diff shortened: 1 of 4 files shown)


GnoX:
@rabits I think you may have missed this PR?

rabits:
Oh yeah sorry, I did - so much things happening)

rabits (approved):
LGTM

PR #50, opened 2025-06-30
Use hass config retry instead of native reconnect logic, improve error handling

GnoX:
This PR adds multiple changes:
- **Reworked reconnection logic** - instead of handling reconnections ourselves, this PR switches control to home assistant config reloads. This is better because it displays errors and retries in integration UI so user does not need to look at logs to see if something failed.
- **Reconnect can now only create one task at a time** - there was possibility that `disconnected` was called more than once which caused multiple reconnect tasks to be spawned at once. If this repeats several times, requests would overlap and device received messages in wrong order causing it to immediately disconnect. This spawned yet more reconnect tasks and existing messages would still be received on the same characteristic but client would not be connected. This would explain what was was happening in #36.
- Added issues in case of connection failures - if the device can not connect more than 10 times in succession, setup fails until it is manually reloaded and creates HA issue that persists until the device is connected successfully
- **Changed update period default to 10 seconds** - it makes more sense to have it a bit higher by default as most of the users don't need it.
- Added observable states to connection for better error reporting
- Added connection timeout option

### DIFF
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -1,19 +1,36 @@
 """The unofficial EcoFlow BLE devices integration"""
 
-from __future__ import annotations
-
 import logging
+from functools import partial
 
+import homeassistant.helpers.issue_registry as ir
 from homeassistant.components import bluetooth
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_ADDRESS, CONF_TYPE, Platform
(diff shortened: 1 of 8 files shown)


rabits:
Thank you! Will quickly check how it's working on my setup and will approve.

rabits (approved):
Looks good. Works for me and defaulting to update per 10 sec is necessary. Geeks will find out how to get immediate updates, but for regular user once per 10 sec sounds reasonable.

PR #52, opened 2025-07-19
Use ruff for linting with official HA rules, add pre-commit rules, add proto definitions

GnoX:
This PR replaces black with ruff and mirrors [official HA rules](https://github.com/home-assistant/core/blob/dev/pyproject.toml#L652) - formatting is mostly the same as black with some very minor changes, but these rules significantly improve code quality. It also adds pre-commit config with ruff and several other handy formatters and CI job that runs it.

PR also adds proto definitions into this repo (will not be installed in HA) to be able to generate typing stubs for development and script that can generates them to correct location - `python -m proto.generate`. While stubs are not required, it greatly improves dev experience (my implementation relies heavily on having these stubs for field typing).


rabits:
Thank you @GnoX - that will be a great addition! Don't you mind if we will pass the SHP2 feature first and then apply new validation rules? Don't want to put additional pressure on him...

GnoX:
No problem, that's what I thought I'd do as well, I can resolve any linting issues really quickly.

This Saturday I'm leaving for a 2 week bike trip though so these changes will have to wait until then, I won't be able to respond during that time.

GnoX:
@rabits Okay, I'm back. Can we look at merging this first as I took over the SHP2 PR?

rabits:
Yeah since SHP2 will take some time I think it's a good idea)

.github/workflows/validate-code-style.yaml
```diff
@@ -2,19 +2,15 @@ name: Validate Code Style
 
 on:
   push:
+    branches: [main]
```
rabits, line 5: Wait, why we verifying on main only? PR's need to be validated as well...

.github/workflows/validate-code-style.yaml
```diff
@@ -2,19 +2,15 @@ name: Validate Code Style
...
-
       - uses: actions/setup-python@v5
         with:
           python-version: "3.x"
-
-      - run: python3 -m pip install black
-
-      - name: Black style validation
-        run: black --check .
+      - uses: pre-commit/action@v3.0.1
```
rabits, line 16: Maybe we can bind it to hash and mark version/date in comment? It's quite out of github-managed repos, so I'm worrying it's prone to tag-replacement attack...

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
```
rabits, line 3: Probably the same here - tags are ok if we have lock file with locked hashes, but I don't see it around, so could you please check if we can use git hash instead and mark tag and date in comment? And the same in the other repos.

proto/generate.py
```diff
@@ -0,0 +1,38 @@
+"""
```
rabits, line 1: Hmmm, in general I like the generation, but removing of generated logic from the repo is not a good idea. The issue is - generation phase means code execution and should be optional for anyone who just wants to use the release from sources. So let's not remove the generated code and keep it in repo - this will help not just folks who don't have protoc, but also us to track changes in generated logic.

rabits (changes requested):
Woops, checked second time and found some concerns, @GnoX could you please check:

.github/workflows/validate-code-style.yaml
GnoX: It does run in PR's, that's what the next line does and you can see it did run in this PR.

proto/generate.py
GnoX: Generated protobufs were not removed in this PR, these are staying in. This script is for updating generated protobufs and generating dev only stubs which are not required for runtime. It only simplifies step of manually compiling them with protoc and copying them over here.

.github/workflows/validate-code-style.yaml
GnoX: Is this a concern in the context of this repo? Is there anything else for attacker to gain other than secrets (which we don't use)?

.github/workflows/validate-code-style.yaml
rabits: Yeah you right, I see now. Weird why I had it for push as well - probably to cover the other branches, but we don't use them so it's fine.

proto/generate.py
rabits: Ah I see now, sorry looked at the v4/v5 removal and did not noticed the other ones...

.github/workflows/validate-code-style.yaml
rabits: Hmmm, in practice we're not modifying the repo during validation and not producing any artifacts, so probably my concern is is too general) Agree, no need for strict measures right away.

.pre-commit-config.yaml
rabits: We don't need to worry about the strict versions because CI is not producing any artifacts and has no access to repo modification, so I think we good.

rabits (approved):
Looks good for sure)

rabits:
Just tested - works nice

PR #53, opened 2025-07-20
Add support for STREAM devices

GnoX:
This PR adds support for STREAM AC, AC Pro, Max, Pro and Ultra, however only AC Pro was tested, see https://github.com/GnoX/ha-ef-ble/issues/16. Untested devices do use the same protobufs but some fields may be unchanged, this will be an easy fix though.


GnoX:
This includes changes from #52 so I'll only take it out of draft after it's merged.

If someone finds this before it is merged and wants to test it, it is available as pre-release [v0.17.0-b.2](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.17.0-b.2) in my fork.

MarjovanLier:
@rabits and @GnoX, I've checked this with my Ultra devices, and it seems to work. I only have two PVs connected, and most of the stats seem accurate compared with what I could try to figure out.

I would just round the PVs with a lambda, like with other devices, else the floats become ints.

GnoX:
@MarjovanLier Great, thanks for testing. 

> I would just round the PVs with a lambda, like with other devices, else the floats become ints.

I only had messages without PV data so I just left them as they are. You're saying that they are currently displayed as ints? Would you be able to paste raw messages with non zero PV data here? You can find raw messages in logs after you enable `Log deserialized device update messages`.

MarjovanLier:
@GnoX sorry for the delay, i created on PR to your repo, https://github.com/GnoX/ha-ef-ble/pull/19

I'll play more on mine :) I have an AC Pro and 2x Ultras

GnoX:
@MarjovanLier I think I addressed issues about rounding and fixed some bugs, can you check if [v0.17.0-b.3](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.17.0-b.3) works correctly?

rabits (approved):
Looks like a very nice addition! Please tell if it's ready - I can merge it in no time.

GnoX:
Ready from my side, feel free to merge anytime.

PR #54, opened 2025-07-21
Add individual circuit control switches for EcoFlow SHP2

crogers2287:
- Add circuit state properties (circuit_1 - circuit_12) to track on/off status
- Add enable methods (enable_circuit_1 - enable_circuit_12) for Home Assistant integration
- Add 12 circuit switch entities with proper outlet device class and icons
- Enhance switch state handling for protobuf enum values (0=OFF, 1=ON)
- Uses existing set_circuit_power() method for reliable BLE communication

This enables individual control of all 12 SHP2 circuits through Home Assistant switches, supporting automation and manual control while maintaining compatibility with the EcoFlow app.

image

PR #56, opened 2025-07-25
Add individual circuit control switches for EcoFlow SHP2

GnoX:
- Add circuit state properties (circuit_1 - circuit_12) to track on/off status
- Add enable methods (enable_circuit_1 - enable_circuit_12) for Home Assistant integration
- Add 12 circuit switch entities with proper outlet device class and icons
- Enhance switch state handling for protobuf enum values (0=OFF, 1=ON)
- Uses existing set_circuit_power() method for reliable BLE communication

This enables individual control of all 12 SHP2 circuits through Home Assistant switches, supporting automation and manual control while maintaining compatibility with the EcoFlow app.

Original author @crogers2287, see #54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+# Python cache files
+__pycache__/
+*.pyc
+*.pyo
+*.pyd
--- a/custom_components/ef_ble/switch.py
+++ b/custom_components/ef_ble/switch.py
@@ -45,78 +45,15 @@ SWITCH_TYPES = [
         icon="mdi:usb",
     ),
     # SHP2 Circuit switches
-    SwitchEntityDescription(
-        key="circuit_1",
-        name="Circuit 1",
-        device_class=SwitchDeviceClass.OUTLET,
-        icon="mdi:power-socket-us",
-    ),
-    SwitchEntityDescription(
-        key="circuit_2",
-        name="Circuit 2",
-        device_class=SwitchDeviceClass.OUTLET,
-        icon="mdi:power-socket-us",
-    ),
-    SwitchEntityDescription(
-        key="circuit_3",
-        name="Circuit 3",
-        device_class=SwitchDeviceClass.OUTLET,
-        icon="mdi:power-socket-us",
-    ),
-    SwitchEntityDescription(
-        key="circuit_4",
-        name="Circuit 4",
-        device_class=SwitchDeviceClass.OUTLET,
-        icon="mdi:power-socket-us",
-    ),
-    SwitchEntityDescription(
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -0,0 +1,5 @@
+# Python cache files
```
rabits, line 1: Yeah, I'm still thinking those are meh - but I think there is no other way if multiple folks will edit the repo...

rabits (changes requested):
Overall it looks good! Tested it on my setup - looks like the circuits are turned off by default... When I try to switch them - they likely return to the original state and sometimes switches - so I'd not say this implementation is anyhow working right now... What I also saw - that by switching one circuit - another is actually triggered (I think it's +1 to the circuit index).

Sorry @GnoX , but it seems will be not that easy to integrate that...

rabits:
Overall I recalled what stops me from implementing circuits for SHP2 - is this advanced logic of combining the circuits into groups: For example I have a panel which uses 240v breakers (takes 2 slots) - it's not a good idea to disconnect just one side of it. And if we start to talk about grouping - that automatically means we need to group the current/power metrics as well... So it's quite a big one actually.

GnoX:
FYI @crogers2287 

@rabits Okay, I also thought it wouldn't be this easy as the data is also incremental which is probably messing up switch states. This is probably similar to the grid power problem we had, maybe I'll try the same fix? 
I wouldn't mind implementing grouping as well, but I'm not really sure how it looks in a data stream, this is not something I would be able to infer from the data I currently have.

rabits:
Oh nono, no worries about grouping - I just thought aloud...

rabits:
Just tested - unfortunately still see the switches as OFF initially... Will try to check what could be wrong there.

custom_components/ef_ble/switch.py
```diff
@@ -58,63 +54,17 @@
...
-        name="Ambient Light",
-    ),
+    # SHP2 Circuit switches
+    *[
+        EcoflowSwitchEntityDescription[shp2.Device](
+            key=f"circuit_{i}",
+            name=f"Circuit {i:02}",
+            device_class=SwitchDeviceClass.OUTLET,
+            icon="mdi:power-socket-us",
+            enable=lambda device, enabled, i=i: device.set_circuit_power(i, enabled),
```
ecammit, line 64: There is an off-by-one issue when sending the command that causes the wrong circuit to be toggled. It's fixed in https://github.com/GnoX/ha-ef-ble/pull/54

GnoX:
Closing in favor of #212

PR #60, opened 2025-08-09
add delta 3 tou / self powered toggles

johnculviner:
Great work on this! It meets 99% of my usecases except for some ability to automate this stuff.

I've added Add toggles for TOU / self powered (or scheduled default).

Thoughts?

Screenshot 2025-08-08 at 10 16 21 PM

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta3.py
+++ b/custom_components/ef_ble/eflib/devices/delta3.py
@@ -119,6 +119,11 @@ class Device(DeviceBase, ProtobufProps):
     )
     dc_charging_current_max = _DcChargingMaxField(pd335_sys_pb2.PV_CHG_VOL_SPEC_12V)
 
+    # Energy strategy operation modes
+    energy_strategy_self_powered = pb_field(pb.energy_strategy_operate_mode, lambda x: x.operate_self_powered_open if x else None)
+    energy_strategy_scheduled = pb_field(pb.energy_strategy_operate_mode, lambda x: x.operate_scheduled_open if x else None)  
+    energy_strategy_tou = pb_field(pb.energy_strategy_operate_mode, lambda x: x.operate_tou_mode_open if x else None)
+
     def __init__(
         self, ble_dev: BLEDevice, adv_data: AdvertisementData, sn: str
     ) -> None:
@@ -268,3 +273,18 @@ class Device(DeviceBase, ProtobufProps):
 
         await self._send_config_packet(config)
         return True
+
+    async def enable_energy_strategy_self_powered(self, enabled: bool):
+        config = pd335_sys_pb2.ConfigWrite()
+        config.cfg_energy_strategy_operate_mode.operate_self_powered_open = enabled
+        await self._send_config_packet(config)
+
+    async def enable_energy_strategy_scheduled(self, enabled: bool):
+        config = pd335_sys_pb2.ConfigWrite()
+        config.cfg_energy_strategy_operate_mode.operate_scheduled_open = enabled
+        await self._send_config_packet(config)
+
+    async def enable_energy_strategy_tou(self, enabled: bool):
+        config = pd335_sys_pb2.ConfigWrite()
+        config.cfg_energy_strategy_operate_mode.operate_tou_mode_open = enabled
+        await self._send_config_packet(config)
--- a/custom_components/ef_ble/switch.py
+++ b/custom_components/ef_ble/switch.py
@@ -44,6 +44,24 @@ SWITCH_TYPES = [
         name="USB Ports",
         icon="mdi:usb",
     ),
+    SwitchEntityDescription(
+        key="energy_strategy_self_powered",
+        name="Self-Powered Mode",
+        device_class=SwitchDeviceClass.SWITCH,
+        icon="mdi:solar-power",
+    ),
+    SwitchEntityDescription(
+        key="energy_strategy_scheduled",
+        name="Scheduled Mode",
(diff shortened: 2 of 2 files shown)


rabits:
Hi @johnculviner , looks nice and simple. Probably we don't need to validate it much - just to fix the style.

joecheve:
Anyone has this added for the delta 3 pro ?  i need to add those  switches  too. thanks

GnoX:
So, I actually thought about adding this initially, but I didn't see too much benefit as TOU can be done with HA automations by lowering the charge limit based on time. The only benefit I see is that it does work even if HA is down. Are there any other benefits being able to set it from HA?

Also I think that `Self-Powered Mode`  switch does the same thing as `Backup Reserve` switch so this one may be redundant.

chopper145:
> So, I actually thought about adding this initially, but I didn't see too much benefit as TOU can be done with HA automations by lowering the charge limit based on time. The only benefit I see is that it does work even if HA is down. Are there any other benefits being able to set it from HA?
> 
> Also I think that `Self-Powered Mode` switch does the same thing as `Backup Reserve` switch so this one may be redundant.

I Use Backup reserve. Low Discharge Value is 15%, Backup Reserve is 30%. But as soon as I do settings around TOU and Self-Powered Mode, it deactivates this Option. Since I use the Unit as a UPS/USV, I do not need the Self-Powered Mode. But I need Backup reserve.


Copied this Text from another Thread:
I would love to have the mode “TOU” to be activated and deactivated with Homeassistant. IT allows me to charge the battery when Solar exceeds my needs from the house. Instead of feeding to grid I would deactivate TOU and so charge the battery until max level reached and then activate it again in the evening to use battery power during night.

Can you add a control switch for mode selects? Or just a switch for TOU?

Thank you.

Best regards
Roger

johnculviner:
hey @rabits sorry for the long delay here. I fixed the style issues, merge conflicts and I think we are good to go!

rabits (approved):
Yep looks nice and clean, thank you!

chopper145:
Hi all
Thank you so much for the toggles. V0.5.9


It works. I am very happy.

I had already crashed my D3P. If you do AC charging speed changes every minute then the EF crashes after around 5-10 minutes. I learned it the hard way since it is our USV/UPS and the whole office went down with NAS, HA, etc...
I am not sure why, because it works with charging active. but does not when charging inactive but sending a command with same or different values every minute.

Anyway. Figured that out and now I have to test the code later on. 


I have it always on Power. AC is always on. I have Tarif set to low price 23.00 to 24.00. I have not connected PV on the EF. But hav it as an installation. Now AI has coded me an automation in Homeassistant to switch TOU mode to off when solar is good. And if it is feeding to grid more and more as the sun rises, the ac charging speed gets set higher and higher. That works pretty well alone and now I have to check it with sun to proof the code.


IF your interested, thats my code:
(Acutally, AI wrote it since I have no idea of programming)

```
alias: EcoFlow Ladeleistung
description: >-
  V5 Stabile PV-Laderegelung: Setzt AC-Charge-Speed nur bei Änderung,
  TOU-sicher, USV-safe
triggers:
  - seconds: /59
    trigger: time_pattern
  - at: "04:00:00"
    trigger: time
actions:
  - choose:
      - conditions:
          - condition: time
            after: "04:00:00"
            before: "04:01:00"
        sequence:
          - target:
              entity_id: switch.ef_dxxxxx_usv_time_of_use_mode
            action: switch.turn_on
          - target:
              entity_id: input_boolean.ecoflow_ladeleistung_automation
            action: input_boolean.turn_off
  - choose:
      - conditions:
          - condition: numeric_state
            entity_id: sensor.pv_uberschuss
            above: 200
        sequence:
          - target:
              entity_id: switch.ef_dxxxxx_usv_time_of_use_mode
            action: switch.turn_off
          - delay:
              minutes: 1
          - target:
              entity_id: input_boolean.ecoflow_ladeleistung_automation
            action: input_boolean.turn_on
  - choose:
      - conditions:
          - condition: template
            value_template: "{{ tou_on and not auto_on }}"
        sequence: []
    default:
      - choose:
          - conditions:
              - condition: template
                value_template: "{{ target_speed | int != last_speed }}"
            sequence:
              - target:
                  entity_id: number.ef_dxxxxx_usv_ac_charging_speed
                data:
                  value: "{{ target_speed }}"
                action: number.set_value
              - target:
                  entity_id: input_number.ecoflow_letzte_ladeleistung
                data:
                  value: "{{ target_speed }}"
                action: input_number.set_value
mode: single
variables:
  tou_on: "{{ is_state('switch.ef_dxxxxx_usv_time_of_use_mode','on') }}"
  auto_on: "{{ is_state('input_boolean.ecoflow_ladeleistung_automation','on') }}"
  soc: "{{ states('sensor.ef_dxxxxx_usv_battery_level') | float(0) }}"
  pv: "{{ states('sensor.pv_uberschuss') | float(0) }}"
  last_speed: "{{ states('input_number.ecoflow_letzte_ladeleistung') | int(0) }}"
  target_speed: |
    {% set reserve = 200 %} {% set soc_limit = 80 %}
    {% if not auto_on %}
      100
    {% elif soc >= 84 %}
      100
    {% elif pv < 200 %}
      100
    {% else %}
      {% set roh = (pv - reserve) * 2 %}
      {% set gestuft = (roh // 100) * 100 %}
      {% set max_leistung = 900 if soc >= soc_limit else 1500 %}
      {{ [100, gestuft, max_leistung] | sort | nth(1) }}
    {% endif %}
```


Have a good day and thank you again.



Cheers
Roger

PR #71, opened 2025-09-08
Add AC port control for SG

### LINKED ISSUE #70: Smart Generator 4000: Switch for AC output
Curious if it would be possible to gain access to the switch for the AC output receptacle on the front of the generator.

GnoX:
It seems that I pluralized field name but left the action method `enable_ac_port` unpluralized causing it to not register as switch.

Fixes #70


rabits (approved):
LGTM

PR #76, opened 2025-09-22
Added a bunch of SHP2 channel fields

rabits:
Finally found a way to add some additional sensors into SHP2.

Had a few issues implementing - thought to use NUM_OF_CHANNELS / NUM_OF_CIRCUITS to unify identical fields but it did not work (HA did not liked those dynamic fields very much and always had issues with type of the field data), so moved static way.

Basically those are working, but still TODO:
* Check PV data input when sun will be back
* Ensure only necessary fields are enabled by default

### DIFF
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -38,46 +38,49 @@ def _create_shp2_binary_sensors():
     # Energy binary sensors
     for i in range(1, shp2.Device.NUM_OF_CHANNELS + 1):
         sensors.update({
-            f"energy{i}_is_enable": BinarySensorEntityDescription(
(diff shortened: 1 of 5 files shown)


rabits:
@GnoX could you please check if I did it correctly and maybe suggest some optimizations?

GnoX:
Everything looks fine to me - I'd probably do it the same way.

It's a bit tedious to name all the fields when it would be very easy to add them in loop though, I'll see if I can do something to make it prettier.

Also pairing fields to sensors is quite a hassle too - it was ok when we had just a few devices but now it's getting out of control. I do have a cool solution to this by auto pairing fields to sensors without coupling library code to HA but its still WIP.

PR #81, opened 2025-10-07
Add support for Delta 3 Classic, add some device specific notes to README

GnoX:
This PR adds support for the Delta 3 Classic which behaves almost exactly like Delta 3 with the exception of usb port control - tested in https://github.com/GnoX/ha-ef-ble/issues/23.

PR also adds some device specific notes to their sections raised in several issues to README.


rabits (approved):
This is very nice change, thank you!

PR #90, opened 2025-10-20
Add STREAM Ultra X to supported devices

### LINKED ISSUE #89: Integration of Ultra stream X
Hello, I'd like to add my ultra stream X but it seems it isn't recognized.
I think not much changes from ultra stream, but the integration says no device supported. How hard would it be to be able to make the ultra stream x be recognized ?
Thanks
Regards

GnoX:
Ultra X has the same I/O as a regular ultra, so this PR only adds serial number prefix to Ultra implementation.

Resolves #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ you have to create it yourself, see this section from the official
 </details>
 
 <details><summary>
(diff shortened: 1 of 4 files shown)


rabits (approved):
Thank you! Nice change

PR #93, opened 2025-11-06
Support for Delta 3 Max Plus

macgaver:
tested and functional on a new Delta 3 Max Plus.
All metrics seems there, it's a clone of the delta3 plus but with updated sn_prefix


GnoX:
Hey, this will work, but a bit cleaner way to do this is to just add serial number prefix to [delta3_plus.py here](https://github.com/rabits/ha-ef-ble/blob/main/custom_components/ef_ble/eflib/devices/delta3_plus.py#L13) and just add this case to [delta3_classic.py here](https://github.com/rabits/ha-ef-ble/blob/main/custom_components/ef_ble/eflib/devices/delta3_classic.py#L134):
```python
case "D3M1":
    model = "Plus Max"
```

Can you also add this device variant to Delta 3 section in readme?

macgaver:
Good review ! Requested changes completed

rabits (approved):
Ok, looks good!) Thank you

Issue #99, opened 2025-11-20
[Bug]: Error trying to add Ultra X


deltamelter:
### What happened?

Running HA Supervised on Raspi 5 and adding new device with latest ha-ef-ble-0.5.4
```
[...]

PR #105, opened 2025-12-02
Regenerate alternator charger protobuf file that could cause a crash

GnoX:
MR for Alternator Charger added protobuf generated with root package which can cause message duplication - reported in #99.


rabits:
@GnoX just a little concern - so you mean you just regenerated those files and they've become different? Could you please tell how that's possible? Is there any changes to proto files was done or generator itself that caused this change in generated files?

GnoX:
What I've done is that I compiled these proto defs before I added package name and forgot to regenerate it again after I did this for dc009 proto. For ge_305, I think I switched proto version to 3 after generation so that's the difference there.

Though it is true that changing version of protoc might serialize them differently. We should probably add a note about a protoc version somewhere or ideally create a dockerfile with fixed environment that generates these.

rabits:
Yeah, I'm usually generate during the build process and looking for differences in the repo, but since we have no build process I think it's ok for now.

Issue #80, opened 2025-10-07
V2 (Delta 2/Wave 2) handshake figured out


hypercorrection:
I have figured out (with a large amount of help from LLMs to deobfuscate code) how to successfully establish a connection with either the Delta 2 or Wave 2 running latest firmware.

The session keys and authentication are working. 

Based on that, this probably works for anything else using the V2 protocol.

The biggest initial difference is in the v2 vs v3 header, the dsrc and ddst fields are missing.

The key derivation seems to be working exactly the same way.

I haven't started digging into the protocol level messages after that, with things like battery level.

But, we are receiving them, and the connection with the device is staying established. If the Wave or Delta don't authenticate, they close the connection within a few seconds. 

I have pushed a version here, which tries to change connect.py as little as possible so it's easier to understand the changeset.

https://github.com/hypercorrection/ef-ble-reverse/commits?author=hypercorrection

I would like to help with getting V2 devices to be supported, and it feels like this is a good starting point.

hypercorrection:
I will share some outputs of the periodic messages once I've given them a once over for identifiers like serials or my user ID. 

I need to get MQTT access via the EF API and I can correlate the values easily too.

GnoX:
Hey, 

that's interesting, if you're able to send me few raw heartbeat messages, I'll probably be able to identify what message parsing code it uses from decompiled app.

Regarding packets - I did in fact try modyfing the code to send v2 headers and someone tested it for me, but it did not work by itself - it did pass initial steps of the auth handshake but got stuck on the last message.

I'll look at the code when I find the time to do so, I've been quite busy these past weeks.

hypercorrection:
@GnoX cool, thank you, I have invited you to a repo with the logs in.

Let me know if/when you need more. 

I've only included the Delta 2 currently, but I think they use the same mh200 module as the base anyway.

GnoX:
Thanks, that looks very promising. While our code cannot decode these messages yet, it will be relatively easy to add once we figure out the format. I'll find some time to look at it more closely the next weekend. 

If you're up to it, you can try to identify the format in the meantime, that would really speed up the implementation.

hypercorrection:
I've updated the repo some more.

I've added MQTT dumps that were taken at the same time as BLE logs for the Wave 2 and Delta 2.

I've also had some luck in getting Claude/Codex to start identifying some of the fields based on the MQTT data.

Hopefully this will save you some more time in figuring out the protocol.

[...]

PR #109, opened 2025-12-07
Add support for v2 packets and initial support for Delta 2

GnoX:
This PR adds support for v2 packets reverse engineered and tested in #80 by @hypercorrection. This opens up possibility to add support for most of the Gen 2 devices although it is will not be straightforward as Gen 3 devices due to Gen 2 devices using fixed width binary encoded messages that are little bit harder to reverse engineer than protocol buffers.

Binary decoding is implemented in `ef_ble.eflib.model.base.RawData` and uses Annotated syntax with python's struct strings as their parameters. Fields are then implemented with the same interface as protobuf fields with RawData messages as their defs.

PR also adds initial support for Delta 2 and Delta 3 1500 (which actually has code name Delta 2 Plus) and few controls.


Gthubln060920:
Should this work with a Delta 2 MAX ?   Version 0.5.6 finds my Delta 2 but not my Delta 2 MAX

GnoX:
> Should this work with a Delta 2 MAX ? Version 0.5.6 finds my Delta 2 but not my Delta 2 MAX

Not yet, this was just for a plain Delta 2. It might work with the Max but I'm going to need some data for that. Feel free to create an issue, we can discuss it there.

Issue #113, opened 2025-12-11
Document Install Using HACS


jkrauska:
Many users (like me) would have benefited from a step by step install flow document showing how to add the repository and configure with HACS.

Also the POST calls and userID data is slightly different now (Dec 2025).  Could use some wiki/documentation refresh.

jkrauska:
If given collaborator access to the wiki, I can assist wit this.

rabits:
Thank you for proposing your help! Let's try that - just invited you. We'll see how it goes.

twl9:
> Also the POST calls and userID data is slightly different now (Dec 2025). Could use some wiki/documentation refresh.

@jkrauska indeed... Maybe you can share a little _sneak preview_, as I tried nummerous times and work with Chrome Console Daily but wasn't yet able to figure out.

@rabits What's also weird is that Ecoflow gives you the option to choose location (US, Europe, etc) on login, but new accounts automaticly only seems to be created for US. So the login option in the latest version gives me 'Login failed: 'Account doesn't exist or incorrect password' what happens when I choose Europe. The add-screen however doesn't allow you to specify location.

introkun:
No wiki was needed on my end. I just added the repo to HACS, downloaded and went through the wizard.

marcuslib:
Have a look here, i did create a video on how to download and install the repo https://www.youtube.com/playlist?list=PLIjDe2oMGO28N6xMz5yzOYJl8u_xUDih4

GnoX:
In #168, we improved readme structure and included the full HACS guide. Integration is now also listed in HACS community store.

We've also included another option to find user ID to the wiki and linked to #130 that has updated manual way to find it.

PR #120, opened 2025-12-13
Add region selector to login screen

GnoX:
We are using `api.ecoflow.com` to get the user id, but it seems that this redirects to the region the user makes request from which should work for most of the users, however in #113 it was reported that the account was created for a different region so this PR adds a selector for region to login to.

It seems that the user id is also not the same depending on which region you choose - I do have accounts for both US and EU regions and it changes based on the endpoint.

### DIFF
--- a/custom_components/ef_ble/config_flow.py
+++ b/custom_components/ef_ble/config_flow.py
@@ -21,7 +21,7 @@ from homeassistant.config_entries import (
     ConfigFlowResult,
     OptionsFlow,
 )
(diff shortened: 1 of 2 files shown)


purplebite:
@GnoX Great contribution! I loved seeing the username/passwork activation option, but had the same experience. When will this be available as new release in HA?

GnoX:
@purplebite As soon as @rabits creates a release - shouldn't take long. If you can't wait to try it out, it was already released in [my fork](https://github.com/GnoX/ha-ef-ble) - you can just install it, get your user id and switch back to this repo.

PR #125, opened 2026-01-06
Add ability to connect to unsupported devices, add diagnostics with packet collection

GnoX:
This PR adds several functional features:
- The main feature is ability to connect to all Ecoflow unsupported devices in order to collect raw packets for adding support for new devices - this was implemented as a new device with `check` method always returning True
- Added diagnostics data for all devices containing basic data by default and added option `Enable packet collection` that includes raw packets in the diagnostics output (disabled by default for supported devices)
- Added descriptions for config options login and region with their explanation

And does some refactoring and internal reorganization:
- Cleaned up schema creation in config flow using builders with pre-defined common options
- Prepared switch to BLE local names instead of generated names
- Added connection event listeners used for observation mainly for diagnostics and logging
- Made Packet return InvalidPacket with error message instead of None for future handling

### DIFF
--- a/custom_components/ef_ble/translations/en.json
+++ b/custom_components/ef_ble/translations/en.json
@@ -62,7 +62,7 @@
       },
       "user": {
         "title": "Add EcoFlow Device",
-        "description": "Devices labeled as [Unsupported] do not expose any sensors but can be used to provide data for implementing support. Refer to the wiki for additional information.",
+        "description": "Devices labeled as **[Unsupported]** do not expose any sensors but can be used to provide data for implementing support. Refer to [this wiki](https://github.com/rabits/ha-ef-ble/wiki/Requesting-Support-for-New-Devices) page for additional information.",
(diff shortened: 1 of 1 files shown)


GnoX:
@rabits Would you mind giving me permission to edit wiki? I'd like to include some info for these features - probably just copying https://github.com/GnoX/ha-ef-ble/wiki/Getting-Diagnostics-Info-for-Supporting-New-Devices over here and adding a link to config flow.

rabits:
Sure man, just sent invite to you - so you will be able to edit wiki.

GnoX:
Thanks, wiki page is now up - https://github.com/rabits/ha-ef-ble/wiki/Requesting-Support-for-New-Devices and linked in the config flow.

rabits:
Ok, will merge this one right after the fix in https://github.com/rabits/ha-ef-ble/issues/42 will be accepted

PR #126, opened 2026-01-08
Fix DP3, Wave3, SG and STREAM devices not resetting updated flag causing cpu increase over time

### LINKED ISSUE #42: [Bug]: CPU usage gradually increases

GnoX:
This PR adds forgotten resets for updated flag causing linear increase in updated fields for each received packet and this is also causing cpu usage to increase in time as number of state updates is also increasing in time.

Should fix #42


rabits (approved):
Great find! Thank you)

PR #127, opened 2026-01-09
Add grid bypass switch to Delta 3

GnoX:
This PR adds grid bypass switch to Delta 3 devices as requested [here](https://community.home-assistant.io/t/ecoflow-ble-unofficial/774794/199).


rabits (approved):
Yep, looks great!)

PR #132, opened 2026-01-14
Dev delta2

chackl1990:
Change from DirectMpptHeartbeatPac to Mr330MpptHeart
Add of DirectBmsMHeartbeatPack for addon battery -  case 0x06, 0x20, 0x32
getting main and slave battery and calculating them like the android app

DEV and TEST not cleaned, and partly AI generated code.

Tried to get set_ac_charging_speed and set_energy_backup_battery_level with looking into app-code

LOG-File of the Delta2 with addon:
[delta2+addon.log.gz](https://github.com/user-attachments/files/24611112/delta2%2Baddon.log.gz)

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta2.py
+++ b/custom_components/ef_ble/eflib/devices/delta2.py
@@ -2,21 +2,40 @@ from ..devicebase import DeviceBase
 from ..model import (
     AllKitDetailData,
     DirectBmsMDeltaHeartbeatPack,
+    DirectBmsMHeartbeatPack,
     DirectEmsDeltaHeartbeatPack,
     DirectInvDelta2HeartbeatPack,
-    DirectMpptHeartbeatPack,
     Mr330MpptHeart,
     Mr330PdHeart,
 )
 from ..packet import Packet
+from ..props import Field
 from ..props.raw_data_field import dataclass_attr_mapper, raw_field
 from ..props.raw_data_props import RawDataProps
 
 pb_pd = dataclass_attr_mapper(Mr330PdHeart)
-pb_mppt = dataclass_attr_mapper(DirectMpptHeartbeatPack)
+pb_mppt = dataclass_attr_mapper(Mr330MpptHeart)
 pb_ems = dataclass_attr_mapper(DirectEmsDeltaHeartbeatPack)
-pb_bms = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
+pb_bms_master = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
+pb_bms_slave = dataclass_attr_mapper(DirectBmsMHeartbeatPack)
 pb_inv = dataclass_attr_mapper(DirectInvDelta2HeartbeatPack)
+# pb = proto_attr_mapper(pd335_sys_pb2.DisplayPropertyUpload)
+
+_PD100_BACKUP_REVERSE_SOC_FIELD = 461
+_PD100_ACP_CHG_POW_MAX_FIELD = 107
+_PD100_ACP_CHG_POW_HAL_MAX_FIELD = 108
+
+_PD100_CFG_BACKUP_REVERSE_SOC_FIELD = 102
+_PD100_CFG_ACP_CHG_POW_MAX_FIELD = 246
+_PD100_CFG_AC_IN_CHG_MODE_FIELD = 125
+
+_PD100_ENERGY_BACKUP_EN_FIELD = 7
+_PD100_CMS_MAX_CHG_SOC_FIELD = 270
+_PD100_CMS_MIN_DSG_SOC_FIELD = 271
+_PD100_CONFIG_READ_ACTION_FIELD = 1
+_PD100_CONFIG_READ_ACTION_DISPLAY = 41
+_PD100_CONFIG_READ_ACTION_BACKUP = 144
+_PD100_AC_IN_CHG_MODE_SELF_DEF_POW = 0
 
 
 class Device(DeviceBase, RawDataProps):
@@ -24,6 +43,8 @@ class Device(DeviceBase, RawDataProps):
 
     SN_PREFIX = (b"R331", b"R335")
     NAME_PREFIX = "EF-R33"
+    _master_bms_msg: DirectBmsMDeltaHeartbeatPack | None
+    _slave_bms_msg: DirectBmsMHeartbeatPack | None
 
     @property
     def packet_version(self):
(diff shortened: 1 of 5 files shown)


chackl1990:
Notes:
set_energy_backup_battery_level
packet = Packet(0x21, 0xFE, 0x11, 0x02, 0x01, value.to_bytes(), version=0x13)

set_ac_charging_speed
packet = Packet(0x21, 0x02, 0xFE, 0x11, 0xB0, 0x03, _encode_varint(value), 0xE8, 0x07, 0x00)

both do not work - i looked them up in the app and did found this - and also checked with AI.
But it is wrong. I checked several hours yesterday with no luck.

Do you have any hints? Maybe i got something wrong in the android app - eg version=0x13 that was the only thing i found in the app and also an AI did not find it.

Also tried Version like on the dc12v (_dc_12v_dst) but it seems only to be applied on the 12v command.

GnoX:
I did map all commands somewhere, but I did skip product type branches. I'll do it again withthem, probably this or next weekend.

> packet = Packet(0x21, 0xFE, 0x11, 0x02, 0x01, value.to_bytes(), version=0x13)

Packet version should always be 0x02, that's the main problem in here. Not sure about the second argument as well, I don't think dst can be anything else than 1-5.

> packet = Packet(0x21, 0x02, 0xFE, 0x11, 0xB0, 0x03, _encode_varint(value), 0xE8, 0x07, 0x00)

Here you have the version set to 0, that's wrong as well. Also, I don't remember any commands accepting varints, they were always fixed.

chackl1990:
Ok - This is what i came up with today:
```
    async def set_ac_charging_speed(self, value: int):
        data_raw = b"\x08" + self._encode_varint(value)
        packet = Packet(0x21, 0x35, 0x14, 0x92, data_raw, version=0x02)
        await self._conn.sendPacket(packet)


    async def set_energy_backup_battery_level(self, value: int):
        data_raw = b"\x08" + self._encode_varint(value)
        packet = Packet(0x21, 0x35, 0x20, 0x84, data_raw, version=0x02)
        await self._conn.sendPacket(packet)
```
        
The data was analyced by AI from the apk - as the last time - i also tried to cross-check it with "set_battery_charge_limit_max".
Next i tried with hci-snooping but i did not get over the ecryption-gap.
Next i tried to recompile the app - also with no luck as the app seems to have a signature check anywere that is activated if you press on agree therms on the login screen.

sorry for the noob question - but how did you reverse engineer the command that the app sends to the ef-device?

kind regards

GnoX:
Hey, sorry, did not find enough time this weekend for this. 

> Ok - This is what i came up with today:

Do those commands work? Something does not look right there.

> sorry for the noob question - but how did you reverse engineer the command that the app sends to the ef-device?

The same way as you did - static analysis of the apk. I found my notes about the commands, quickly looked through the app and tried to create a command table that might help you. Not guaranteed that everything is correct though.

For all commands
`dst` = 0x21
`cmd_set` = 0x20

| Method Name                 |          src | cmd_id | Payload                                          |
|-----------------------------|-------------:|-------:|--------------------------------------------------|
| `set_charge_upper_limit`    |         0x03 |   0x31 | `[limit (U8)]`                                   |
| `set_discharge_lower_limit` |         0x03 |   0x33 | `[limit (U8)]`                                   |
| `set_xboost`                | 0x04 or 0x05 |   0x42 | `[0xFF, on=0,1 (U8), 0xFF,0xFF,0xFF,0xFF,0xFF]`  |
| `set_energy_backup`         |         0x02 |   0x5E | `[enabled=0,1 (U8), bp_soc (U8), 0x00, 0x00]`    |
| `enable_solar_priority`     |         0x02 |   0x5C | `[on (U8)]`                                      |
| `set_ac_frequency`          | 0x04 or 0x05 |   0x42 | `[0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,frequency (U8)]` |
| `set_ac_charge_power_mr330` |         0x05 |   0x45 | `[watts (LE16), 0xFF]`                           |
| `set_ac_charge_power_mr350` |         0x04 |   0x45 | `[0xFF,0xFF, watts (LE16), 0xFF]`                |
| `set_ac_auto_on`            |         0x04 |   0x54 | `[on (U8)]`                                      |
| `set_dc_output_switch`      | 0x05 or 0x07 |   0x51 | `[on (U8)]`                                      |
| `set_car_charge_current`    |         0x05 |   0x47 | `[in_current (LE32), out_current (LE32)]`        |
| `set_usb_output_switch`     |         0x02 |   0x22 | `[on (U8)]`                                      |
| `set_screen_brightness`     |         0x02 |   0x27 | `[0xFF, 0xFF, brightness (U8)]`                  |

There's more commands there, I only quickly went through the ones I recognized quickly.

chackl1990:
Thank you, ill take look tomorrow.
My cmds did not work but ill try your list tomorrow and try to find those in the app.

chackl1990:
```
    async def set_ac_charging_speed(self, value: int):
        payload = self._to_le16(value) + bytes([0xff])
        packet = Packet(0x21, 0x05, 0x20, 0x45, payload, version=0x02)
        await self._conn.sendPacket(packet)

    async def set_energy_backup_battery_level(self, value: int):
        payload = bytes([0x01]) + value.to_bytes() + bytes([0x00, 0x00])
        packet = Packet(0x21, 0x02, 0x20, 0x5E, payload, version=0x02)
        await self._conn.sendPacket(packet)
```

this worked now on my version :)

i'll take a detailed look in the version the next days and in testing how "save" the bt compared to the ecoflow-cloud package is.

Thank you for your input - it realy helped me!
Question: do you have a delta with a other version so it can be tested?

WHY: During my test i managed to set the ch-speed to something like 45000+ Watts ...
OK i thought that would not be a problem i was save an unpluugged the delta. Safe becaue if i checked mqtt over ecoflow-cloud i was seeing the 45000+ Watts.

So just to be shure that nothing get damaged a test would be fine.

GnoX:
> Question: do you have a delta with a other version so it can be tested?

No, I only own two gen 3 devices. We can ask someone in #80 to test it.

> WHY: During my test i managed to set the ch-speed to something like 45000+ Watts 

You may have mistakenly used incorrect encoding. I don't know how your `_to_le16` looks, but if you do something like
`160.to_bytes(2)`  this would be 40960 in little endian - `to_bytes` uses big endian by default, you have to use `to_bytes(2, "little")` to get it right.

> OK i thought that would not be a problem i was save an unpluugged the delta. Safe becaue if i checked mqtt over ecoflow-cloud i was seeing the 45000+ Watts.

Hardware would limit it anyways, it won't allow you to draw more than it's designed for.


Also, would you mind if, after you get everything working, I'd take over your implementation and clean it up a little? It's quite hard to read in the current state and I think it would take more time for me to point all of it out than to just change it myself.

chackl1990:
yes i'll do that - I think it is better closing this as i design the code new - but i will need till weekend and further testing.

I would first adapt the delta2.py - i need some sensors like capacity (remain) to calulate the battery state and i need to test with and without addon battery.

How would you handle when adding new sensors that are not in sensors.py atm? So that the sensor settings can be used also in other devices.

So first stage make delta2.py work with existing entities
Stage 2 add entities that are missing.
How would you handle this?

EDIT:
Question 1 about safety for holding the charge speed in a value range:
about 2 years ago when experimening with the cloud i had set to much chgspeed during an error that pulled the 10amp (230V) fuse the device has - so i never test with ac plugged in in my tests since then. 10amps should be 2300 watts and I maximum pull 200W out of AC bypass during charge. Maximum in app is 1500W + 200W online but my fuse went off. Setting it to 1500W would not blow it.

Question 2 about memory writes:
The delta2 saves the values on (i think flash) memory - i wrote EF during the implimentation of the public API to ask how much it is save to change. They said that there is a write load but did not say how much would be safe.
But to be shure to ask: There is no need to desing something to prevent damage of the flash memory?

Kind regards

GnoX:
> How would you handle when adding new sensors that are not in sensors.py atm? So that the sensor settings can be used also in other devices.

I'm not sure what you mean by this. Did you not already add some new sensors there? The key of that dictionary in sensors.py corresponds to the field name in the device - that's how it's added as a sensor.

> The delta2 saves the values on (i think flash) memory - i wrote EF during the implimentation of the public API to ask how much it is save to change. They said that there is a write load but did not say how much would be safe.
But to be shure to ask: There is no need to desing something to prevent damage of the flash memory?

I don't think you need to worry about this, it's not like it's writing gigabytes of data there. What we're doing here is nothing different than what the app is doing.

chackl1990:
Ok thank you! I'm already testing the addon battery + ac-speed + backup-level - i'll commit in an seperate request and close this draft.

PR #143, opened 2026-01-23
Additions for Delta2

chackl1990:
Add Battery-Slave-Data DirectBmsMHeartbeatPack
Add Battery-Level Master + Slave based on calculations in app
Add AC Charge Speed
Add Energy Backup Level

Tested on device that is not product-type 82.
Maybe if someone can give me feedback with product-type 82
I do not know if it is device age or just testing vom EF....

Enhance Product version Handling (_product_type_82)

**need your feedback:**
i implemented new sensors.

there are missing switches, buttons and numbers (like xboost that were not used up to now in other dervices)
is it ok to go for those like that or should i keep in mind that some devices can have 2 addon batterys so we maybe need a battery_level_slave_1 and battery_level_slave_2, i keept now with battery_level_slave

I just wanted to know how you like to define those entities.

Kind regards

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta2.py
+++ b/custom_components/ef_ble/eflib/devices/delta2.py
@@ -17,15 +17,11 @@ class _BmsHeartbeatBattery1(DirectBmsMDeltaHeartbeatPack):
     pass
 
 
-class _BmsHeartbeatBattery2(DirectBmsMDeltaHeartbeatPack):
-    pass
-
-
 pb_pd = dataclass_attr_mapper(Mr330PdHeart)
 pb_mppt = dataclass_attr_mapper(Mr330MpptHeart)
 pb_ems = dataclass_attr_mapper(DirectEmsDeltaHeartbeatPack)
-pb_bms_master = dataclass_attr_mapper(_BmsHeartbeatBattery1)
-pb_bms_slave = dataclass_attr_mapper(_BmsHeartbeatBattery2)
+pb_bms = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
+pb_bms_1 = dataclass_attr_mapper(_BmsHeartbeatBattery1)
 pb_inv = dataclass_attr_mapper(DirectInvDelta2HeartbeatPack)
 
 
@@ -43,19 +39,17 @@ class Device(DeviceBase, RawDataProps):
     ac_input_power = raw_field(pb_pd.ac_input_watts)
     plugged_in_ac = raw_field(pb_pd.ac_charge_flag, lambda x: x == 1)
 
-    battery_level_main = raw_field(pb_bms_master.f32_show_soc, lambda x: round(x, 2))
-    battery_1_battery_level = raw_field(
-        pb_bms_slave.f32_show_soc, lambda x: round(x, 2)
-    )
+    battery_level_main = raw_field(pb_bms.f32_show_soc, lambda x: round(x, 2))
+    battery_1_battery_level = raw_field(pb_bms_1.f32_show_soc, lambda x: round(x, 2))
     # battery_level = Field[float]()
     battery_level = raw_field(pb_ems.f32_lcd_show_soc, lambda x: round(x, 2))
 
-    master_design_cap = raw_field(pb_bms_master.design_cap)
-    master_remain_cap = raw_field(pb_bms_master.remain_cap)
-    master_full_cap = raw_field(pb_bms_master.full_cap)
-    slave_design_cap = raw_field(pb_bms_slave.design_cap)
-    slave_remain_cap = raw_field(pb_bms_slave.remain_cap)
-    slave_full_cap = raw_field(pb_bms_slave.full_cap)
+    master_design_cap = raw_field(pb_bms.design_cap)
+    master_remain_cap = raw_field(pb_bms.remain_cap)
+    master_full_cap = raw_field(pb_bms.full_cap)
+    slave_design_cap = raw_field(pb_bms_1.design_cap)
+    slave_remain_cap = raw_field(pb_bms_1.remain_cap)
+    slave_full_cap = raw_field(pb_bms_1.full_cap)
     battery_addon = Field[bool]()
 
     energy_backup_battery_level = raw_field(pb_pd.bp_power_soc)
@@ -123,10 +117,10 @@ class Device(DeviceBase, RawDataProps):
                 self.update_from_bytes(DirectEmsDeltaHeartbeatPack, packet.payload)
                 processed = True
             case 0x03, 0x20, 0x32:
-                self.update_from_bytes(_BmsHeartbeatBattery1, packet.payload)
+                self.update_from_bytes(DirectBmsMDeltaHeartbeatPack, packet.payload)
                 processed = True
             case 0x06, 0x20, 0x32:
-                self.update_from_bytes(_BmsHeartbeatBattery2, packet.payload)
+                self.update_from_bytes(_BmsHeartbeatBattery1, packet.payload)
                 processed = True
             case 0x04, _, 0x02:
                 self.update_from_bytes(DirectInvDelta2HeartbeatPack, packet.payload)


custom_components/ef_ble/sensor.py
```diff
@@ -256,6 +256,49 @@ def _wave_unit(dev: wave3.Device):
...
+        native_unit_of_measurement="mAh",
+        state_class=SensorStateClass.MEASUREMENT,
+        entity_registry_enabled_default=False,
+    ),
+    "slave_design_cap": SensorEntityDescription(
+        key="slave_design_cap",
+        native_unit_of_measurement="mAh",
+        state_class=SensorStateClass.MEASUREMENT,
+        entity_registry_enabled_default=False,
+    ),
```
GnoX: Is this battery capacity useful in any automations? I intentionally didn't add any of these as sensors as this is not very useful information and it is static. If you need it, I would add it as an attribute to `battery_level`.

custom_components/ef_ble/eflib/model/direct_bms_heartbeat_pack.py
```diff
@@ -32,3 +32,7 @@ class DirectBmsMDeltaHeartbeatPack(RawData):
     input_watts: Annotated[int, "I", "inputWatts"]
     output_watts: Annotated[int, "I", "outputWatts"]
     remain_time: Annotated[int, "I", "remainTime"]
+
+
+class DirectBmsMHeartbeatPack(DirectBmsMDeltaHeartbeatPack):
+    pass
```
GnoX: What's the purpose of this class? Does this need to be here if it has no fields?

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -141,5 +200,37 @@ def _update_product_type(self, detail: AllKitDetailData) -> None:
...
+            total_remain = (master_remain or 0) + (slave_remain or 0)
+            self.battery_level = round((total_remain / total_full) * 100, 2)
+        else:
+            self.battery_level = self.battery_level_main
+
+    @staticmethod
+    def _to_le16(value: int) -> bytes:
+        # clamp to 0..65535
+        value = min(max(value, 0), 0xFFFF)
+        return value.to_bytes(2, byteorder="little", signed=False)
```
GnoX: I don't think these need to be clamped, it would be a bug if  ac charging speed would receive values higher than this. It should rather fail, not clamp it. Just use `to_bytes` directly in the payload.

Just remove this method and handle it directly in the set method.

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -114,6 +143,36 @@ async def set_battery_charge_limit_min(self, limit: int):
...
+        if self.max_ac_charging_power is None:
+            return
+
+        value = max(1, min(value, self.max_ac_charging_power))
+        # Sending 0 sets to (more than) max-load - better safe
+
+        if self._product_type_82():
+            payload = bytes([0xFF, 0xFF]) + self._to_le16(value) + bytes([0xFF])
+        else:
+            payload = self._to_le16(value) + bytes([0xFF])
```
GnoX: This would look better like this


```suggestion
        value = value.to_bytes(2, "little")
        payload = bytes([ value, 0xFF])
        if self._product_type_82():
            payload = bytes([0xFF, 0xFF]) + payload
```

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -114,6 +143,36 @@ async def set_battery_charge_limit_min(self, limit: int):
...
+            self.battery_charge_limit_min is None
+            or self.battery_charge_limit_max is None
+        ):
+            return
+
+        value = max(
+            self.battery_charge_limit_min,
+            min(value, self.battery_charge_limit_max),
+        )
+        payload = bytes([0x01]) + value.to_bytes() + bytes([0x00, 0x00])
```
GnoX, line 184:
```suggestion
        payload = bytes([0x01, value, 0x00, 0x00])
```

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -141,5 +200,37 @@ def _update_product_type(self, detail: AllKitDetailData) -> None:
...
         self._product_type = detail.kit_base_info[0].product_type
 
+    def _update_ac_chg_limits(self) -> None:
+        if self.battery_addon:
+            self.max_ac_charging_power = 1500
+        else:
+            self.max_ac_charging_power = 1200
+
+    def _product_type_82(self) -> bool:
+        return self._product_type == 82
```
GnoX: Rename this to `_is_mr530` to be more explicit about what that product type represents. That's how it's named in the app.

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -141,5 +200,37 @@ def _update_product_type(self, detail: AllKitDetailData) -> None:
...
+
+    def _ac_speed_dst(self) -> int:
+        return 0x04 if self._product_type_82() else 0x05
```
GnoX: I don't think this needs to be as a separate method, just use it directly in the set.

GnoX:
> there are missing switches, buttons and numbers (like xboost that were not used up to now in other dervices)
is it ok to go for those like that or should i keep in mind that some devices can have 2 addon batterys so we maybe need a battery_level_slave_1 and battery_level_slave_2, i keept now with battery_level_slave
> 
> I just wanted to know how you like to define those entities.

Oh, is this what you meant by that sensor question in the previous MR? You can just use fields named `battery_<index>_battery_level` and they would be added automatically for every battery (disabled by default).

custom_components/ef_ble/eflib/model/direct_bms_heartbeat_pack.py
chackl1990: i think DirectBmsMHeartbeatPack was something i saw in the app when reading the extra-battery.
Since the class is the same as DirectBmsMDeltaHeartbeatPack i think this is not needed - i'll test it.

custom_components/ef_ble/sensor.py
chackl1990: i wanted to use them for future soh states - but yes, not realy that needed so i put them on default disabled.
should i remove them?

custom_components/ef_ble/eflib/model/direct_bms_heartbeat_pack.py
GnoX: I see that it's the other way around - `DirectBmsMDeltaHeartbeatPack` extends `DirectBmsMHeartbeatPack`, not the way you defined it here. The base class is in fact empty and mainly used for instance checking - we don't need to do that so I chose to ignore it and implemented the delta specific one without inheriting from this one.

custom_components/ef_ble/eflib/devices/delta2.py
chackl1990: payload = bytes([0x01]) + value.to_bytes() + bytes([0x00, 0x00])
payload = bytes([0x01, value.to_bytes(), 0x00, 0x00])

need to thest this - i had a little struggle there as i am not fixed here :)
Thank you for your suggestion

custom_components/ef_ble/sensor.py
GnoX: Wouldn't you be able to use it for soh if they weren't sensors? You can keep it as a field in device, they would just not be exposed as sensors.

I wouldn't mind exposing them though, it's just that these are constants, right? It would be better to use attributes for that.

custom_components/ef_ble/eflib/devices/delta2.py
GnoX: Oops, I screwed up in the suggestion, it should be
```python
        payload = value + bytes([0xFF])
```

custom_components/ef_ble/sensor.py
chackl1990: i made the sensors to check the value - i just left them there if someone whats to enable them. I made a mistrake first an i calculated the battery by using not the full_cap but the desing_cap.

so full_cap changes slowly down - design_cap is fixed per device.
full_cap is needed for battery calculations.

Do you have an example for how to set them as constant?

custom_components/ef_ble/eflib/devices/delta2.py
GnoX: Sorry, I thought I edited it before you noticed - there shouldn't be `to_bytes` in the list.

custom_components/ef_ble/sensor.py
GnoX: Ah, I just understood what you mean - you want these to create helpers that calculate remaining capacity like `remaining_capacity = capacity * soh`, right?

To add them as attributes, you can add them to `battery_level` like this
```python
    "battery_level": EcoflowSensorEntityDescription(
        key="battery_level",
        native_unit_of_measurement=PERCENTAGE,
        device_class=SensorDeviceClass.BATTERY,
        state_class=SensorStateClass.MEASUREMENT,
        state_attribute_fields=["master_full_cap", "slave_design_cap"],
    ),
```

custom_components/ef_ble/eflib/devices/delta2.py
chackl1990: ok should i also remove the _dc_12v_dst as made in the fixes?

GnoX: Yes, it's a simple conditional that's only used once in the set ac power method so it doesn't really make sense to have it as a method.

custom_components/ef_ble/sensor.py
chackl1990: ok, what happens in other devices if there is no master_full_cap or slave_design_cap? (am i right that eg master_full_cap is the class variable field of the device?)
Thank you

GnoX: If device class does not define these fields, it just won't add them.

chackl1990:
> > there are missing switches, buttons and numbers (like xboost that were not used up to now in other dervices)
> > is it ok to go for those like that or should i keep in mind that some devices can have 2 addon batterys so we maybe need a battery_level_slave_1 and battery_level_slave_2, i keept now with battery_level_slave
> > I just wanted to know how you like to define those entities.
> 
> Oh, is this what you meant by that sensor question in the previous MR? You can just use fields named `battery_<index>_battery_level` and they would be added automatically for every battery (disabled by default).

yes - as i saw that it is designed to be reused in other devices.
So i should change 'battery_level_slave' to 'battery_1_battery_level'?

GnoX:
> yes - as i saw that it is designed to be reused in other devices.
> So i should change 'battery_level_slave' to 'battery_1_battery_level'?

Yeah, let's make it consistent.

chackl1990:
Ok - did you already design a full calculation of soc with multible batteries?

EDIT:
and there was a battery_level and a battery_level_main in the sensors-list - so i used them - are those correct?

GnoX:
> Ok - did you already design a full calculation of soc with multible batteries?

No, Gen 3 battery devices have these fields already calculated. I didn't have any samples with multiple batteries so I could not verify if D2 does not have field like this. Check if `soc` in `DirectEmsDeltaHeartbeatPack` message doesn't already do this.

custom_components/ef_ble/eflib/model/direct_bms_heartbeat_pack.py
chackl1990: So this is what i have now:
pb_bms_master = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
pb_bms_slave = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)

and now her:
            case 0x03, 0x20, 0x32:
                self.update_from_bytes(DirectBmsMDeltaHeartbeatPack, packet.payload)
                processed = True
            case 0x06, 0x20, 0x32:
                self.update_from_bytes(DirectBmsMDeltaHeartbeatPack, packet.payload)
                processed = True
                
Now my battery level jump between the slave and main one - i would need a 2nd class here with another name.
or is it possible to revere to pb_bms_master and pb_bms_slave?

GnoX: Ooh, right, that was the reason you used that empty class. I designed it for one message per field. Okay, what you can temporarily do is define classes in `delta2.py` that look like this:
```python
class _BmsHeartbeatBattery1(DirectBmsMDeltaHeartbeatPack):
    pass

class _BmsHeartbeatBattery2(DirectBmsMDeltaHeartbeatPack):
    pass

...

pb_bms_1 = dataclass_attr_mapper(_BmsHeartbeatBattery1)
...
```

I'll redesign it later.

chackl1990:
> > Ok - did you already design a full calculation of soc with multible batteries?
> 
> No, Gen 3 battery devices have these fields already calculated. I didn't have any samples with multiple batteries so I could not verify if D2 does not have field like this. Check if `soc` in `DirectEmsDeltaHeartbeatPack` message doesn't already do this.

yep it has it there! - so switching to this - and so also alle the caps are not needet for the moment - so i'll put them away to get this to finish

chackl1990:
setting as draft - as there ist an open problem with using 2 times the Class DirectBmsMDeltaHeartbeatPack

custom_components/ef_ble/translations/en.json
```diff
@@ -279,6 +279,27 @@
...
+      },
+      "slave_design_cap": {
+        "name": "Slave Battery Design Capacity"
+      },
+      "slave_remain_cap": {
+        "name": "Slave Battery Remaining Capacity"
+      },
+      "slave_full_cap": {
+        "name": "Slave Battery Full Capacity"
+      },
```
GnoX: You removed the sensors but didn't remove translations for them.

chackl1990:
can you pls check this there:

        value = value.to_bytes(2, "little")
        payload = bytes([value, 0xFF])
        if self._is_mr530():
            payload = bytes([0xFF, 0xFF]) + payload
            
we try to get bytes from what is already bytes. 'bytes' object cannot be interpreted as an integer
i would just fallback to a string combine solution i had before - do you have a better idea?

GnoX:
> can you pls check this there:
> 
> ```
>     value = value.to_bytes(2, "little")
>     payload = bytes([value, 0xFF])
>     if self._is_mr530():
>         payload = bytes([0xFF, 0xFF]) + payload
> ```
> 
> we try to get bytes from what is already bytes. 'bytes' object cannot be interpreted as an integer i would just fallback to a string combine solution i had before - do you have a better idea?

Check my comment for that line, I realized I did it wrong and fixed in the next one.

Just change second line to `payload = value + bytes([0xFF])`

chackl1990:
ok i hope i did not forget something if so than sorry pls let me know.

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -8,13 +8,24 @@
...
+class _BmsHeartbeatBattery2(DirectBmsMDeltaHeartbeatPack):
+    pass
+
+
 pb_pd = dataclass_attr_mapper(Mr330PdHeart)
 pb_mppt = dataclass_attr_mapper(Mr330MpptHeart)
 pb_ems = dataclass_attr_mapper(DirectEmsDeltaHeartbeatPack)
-pb_bms = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
+pb_bms_master = dataclass_attr_mapper(_BmsHeartbeatBattery1)
+pb_bms_slave = dataclass_attr_mapper(_BmsHeartbeatBattery2)
```
GnoX: I meant to use this indices only for individual batteries, not for the main one, for main one, use the original message. It should ideally look like this

```suggestion
pb_bms = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
pb_bms_1 = dataclass_attr_mapper(_BmsHeartbeatBattery1)
pb_bms_2 = dataclass_attr_mapper(_BmsHeartbeatBattery2)
```

chackl1990: ok - the delta2 can add one additional battery not more so i do this:
pb_bms = dataclass_attr_mapper(DirectBmsMDeltaHeartbeatPack)
pb_bms_1 = dataclass_attr_mapper(_BmsHeartbeatBattery1)

use DirectBmsMDeltaHeartbeatPack for the main and use _BmsHeartbeatBattery1 for addon correct?

GnoX: Ah, thought you could chain more of them. Yes, that's right, just ignore 2.

chackl1990:
I have to say thank you for your patience with me - since i made only a few commits by just changing just some lines.
kind regards

GnoX:
Acutally there's one more thing - I don't think that that product type 82 is used for standard delta 2s. I was looking in the app and was finally able to associate product types to the actual devices - 82 is delta pro. I think I just initially chose incorrect headers for product type 82 and noone tested those particular commands and that's where I confused you. Also, from delta 2 logs I had, the AllKitDetailData structure was always filled with zeros so the product type would not work there either. In app, product type is most of the time determined from serial number prefixes.

We should just remove the product type logic entirely and just use the correct headers.

chackl1990:
ok - should i change it?
i just can do it in about 6 hours when i am home.

GnoX:
It'll be faster if I did it along with some cleanup so don't worry about it - it does not affect functionality so it can be done later.

chackl1990:
ok just tell me if and when i should change it.
If you'd do a cleanup an need a test just tell me - i'll test it with my d2.

PR #147, opened 2026-01-31
Add support for Smart Meter

GnoX:
This PR adds support for Smart Meter - sensors tested in https://github.com/GnoX/ha-ef-ble/issues/41

Also takes out link from translation to placeholder which fixes hassfest CI check.


rabits (approved):
Yeah looks good!

PR #148, opened 2026-02-01
Add main battery level to STREAM devices

### LINKED ISSUE #101: [Bug]: two Ecoflow Stream AC only one SOC and Load from Battery
### What happened?

I have two Ecoflow Stream AC Pro units. They are connected in different rooms.

In the Ecoflow app, I have to set them up under the same system; otherwise, I can only assign my Tibber Pulse to one of the two units.

The HA-EF-Ble connection works well, but only the state of charge (SOC) of one unit is displayed; the other shows 0. I can see them separately in the app.

Image

When I turn off one device, the SoC switches to the other device.

Image

When I turn the device back on, it remains on the second device.

Image

Version 0.5.3

both

USB dongle

_No response_

```text

```

### Relevant log output

```shell

```

GnoX:
This PR is very small change that resolves #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,10 +199,11 @@ you have to create it yourself, see this section from the official
 | *Sensors*                   | *Switches*  | *Sliders*             | *Selects*       |
 |-----------------------------|-------------|-----------------------|-----------------|
 | Battery Level               | Feed Grid   | Feed Grid Power Limit | Energy Strategy |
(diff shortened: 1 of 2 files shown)


GnoX:
For for the CI job is in #147 so check that one first.

rabits (approved):
Yep thank you, looks great!

PR #157, opened 2026-02-07
Set max AC charging power to 1000W for Delta 3 2000 Air

JMacalinao:
The Delta 3 2000 Air supports AC charging speeds of up to 1000W. This PR sets the max to match the spec for this model.

I also fixed the conditions in the device property so the device model shows up properly in HA.

Both fixes are tested using my own device and looks to be working properly.


GnoX:
Oh, I missed this, thanks.

PR #160, opened 2026-02-07
Add initial support for Stream Microinverter

### LINKED ISSUE #84: Can the Ecoflow STREAM Microinverter be added?
Just curious if the stream microinverter could be added and controlled by this integration? 
Any thoughts or information would be great. 

We just added the alternator charger using this to HA and it's incredible.

GnoX:
This PR adds initial support for Stream Microinverter. As Microinverter is relying on interruptable input, it will try to reconfigure each time it is disconnected. It would be better if it it showed status as Off instead of going into reconfigure mode - this may get addressed later.

Resolves #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,25 @@ you have to create it yourself, see this section from the official
 </details>
 
 <details><summary>
+<b>STREAM Microinverter</b>
+</summary>
+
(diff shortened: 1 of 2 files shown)


rabits (approved):
Perfect as usual!) Thank you!

PR #181, opened 2026-02-17
Add solar input and 12V DC output sensors for Delta 2

neotigger:


custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -21,7 +21,9 @@ class Device(Delta2Base):
     energy_backup_battery_level = raw_field(pb_pd.bp_power_soc)
     dc_output_power = raw_field(pb_pd.dc_pv_output_watts)
     ac_charging_speed = raw_field(pb_mppt.cfg_chg_watts)
-
+    solar_input_power = raw_field(pb_pd.dc_pv_input_watts)
+    dc_car_out_power = raw_field(pb_pd.car_watts)
```
GnoX, line 25: If you name this field `dc12v_output_power`, you don't have to add `dc_car_out_power` in `sensors.py`

```suggestion
    dc12v_output_power = raw_field(pb_pd.car_watts)
```

custom_components/ef_ble/eflib/devices/delta2.py
```diff
@@ -21,7 +21,9 @@ class Device(Delta2Base):
     energy_backup_battery_level = raw_field(pb_pd.bp_power_soc)
     dc_output_power = raw_field(pb_pd.dc_pv_output_watts)
     ac_charging_speed = raw_field(pb_mppt.cfg_chg_watts)
-
+    solar_input_power = raw_field(pb_pd.dc_pv_input_watts)
```
GnoX, line 24: This is not just solar, right? It is DC input through XT60 port. I would use `dc_port_input_power` for this.

```suggestion
    dc_port_input_power = raw_field(pb_pd.dc_pv_input_watts)
```

custom_components/ef_ble/sensor.py
```diff
@@ -917,6 +917,22 @@ def _wave_unit(dev: wave3.Device):
...
+        entity_registry_enabled_default=True,
+    ),
+    "dc_car_out_power": SensorEntityDescription(
+        key="dc_car_out_power",
+        name="12V DC Output",
+        native_unit_of_measurement=UnitOfPower.WATT,
+        device_class=SensorDeviceClass.POWER,
+        state_class=SensorStateClass.MEASUREMENT,
+        entity_registry_enabled_default=True,
+    ),
```
GnoX, line 935: No need for any of these if you use already existing 12v field. And `solar_input_power` is already registered here.


```suggestion
```

custom_components/ef_ble/eflib/devices/delta2.py
GnoX: Ah, I see that you can charge it through 12V port as well, this would make it confusing.

Issue #176, opened 2026-02-16
Support for Delta 3 with dual AC output switches


scotte:
Some Delta 3 models, such as the Delta 3 Max Plus, have dual AC output switches that are controlled separately. This integration currently only controls and shows one of them.

I'm not sure exactly which models have this feature, but please let me know if I can provide any additional details or testing on my 3 Max Plus.

Thanks for this great integration!

GnoX:
PR already up - you can try it in version [v0.33.0-f.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.33.0-f.1) from my fork.

scotte:
Thanks @GnoX - I didn't see an issue for it and missed the abbreviation of "D3M+".

I tried the branch in your PR at fe1a470 but it doesn't work right - the new entity shows up, but it is always shown as "On", turning it Off immediately switches back to On, and it does not reflect actual device state. I did delete and re-add the integration device to hass. Let me know any additional details I can provide.

[EDIT: fixed commit SHA - pasted in the wrong one previously]

GnoX:
> Thanks @GnoX - I didn't see an issue for it and missed the abbreviation of "D3M+".

You didn't miss anything, I created that PR as a respoonse to this issue.

I noticed one error, can you redownload the same version again?

scotte:
Ah, thanks - your reply was so quick I assumed I had just missed an existing request!

Let me try again - I'll use the released version instead of checking out your fork this time just to be sure I have the right commit/version and will update shortly.

scotte:
@GnoX I used this tarball directly - https://github.com/GnoX/ha-ef-ble/archive/refs/tags/v0.33.0-f.1.tar.gz and the "AC Ports (2)" entity disappeared entirely.

GnoX:
Huh, nothing there looks wrong. Do you see any errors in the logs?

**edit**: Actually, the only way it can disappear is that you accidentally install it using the main integration instead of my fork, maybe? They are indistinguishable from devices screen if you have both of them installed at the same time. I actually never tested what happens in that case, I think one of them would crash.

scotte:
The problem is that both `delta3_plus.py` and `delta3_max_plus.py` match based on `SN_PREFIX` and the same prefix is used for both models, so it's probably luck of the inode in which one gets loaded first that determines which model is detected. I put a bogus value into the model I don't have (delta3_plus) and the second AC output is back now that my device is detected as the right model.

[...]

PR #182, opened 2026-02-17
Add power for individual AC ports and usb C (3) power to D3M+

GnoX:
This PR adds new sensors that are only available for Delta 3 Max Plus - Power for AC Ports 1, 2 and a third USB C port.

Continuation of #176


rabits (approved):
A very nice change!

PR #185, opened 2026-02-22
Add support for PowerStream (encryption type 1), improve packet handling, diagnostics and auth error messages

### LINKED ISSUE #18: Support for EcoFlow powerstream.
Would love to see support for the EcoFlow powerstream. As the app ain't all that helpful in setting the strategies I want. 

If you need some information, let me know and I will see what I can do to provide it

GnoX:
This PR adds several major features, the most important is type 1 encryption and improvement of packet decoding.

Full list of changes:
- initial support for PowerStream
- connection can now recognize that the device uses type 1 encryption from advertisement data and successfully authenticate
- decoding of packets is now more robust - instead of blindly trusting encoded header lengths, it scans for prefixes and can handle false prefixes inside payloads by only advancing if crc16 matches
- improved error messages for config flow that occured during authentication procedure
- used AES encryption for all potentially sensitive info when sharing diagnostics, only developers have access to its private key

Thanks to @MarjovanLier that provided working initial implementation for PowerStream.

Resolves #18

Additional info from @MarjovanLier:

PowerStream uses MD5-derived keys instead of ECDH for BLE encryption. Subclasses Connection with overrides for key derivation, zero-padded AES-CBC encryption, and direct Packet framing (no EncPacket wrapper). Device stub with data_parse returning False — data streaming is a follow-up phase.

Device pushes unsolicited V19 plaintext data before responding to auth, which broke the single-shot getAuthStatus handler. Skip getAuthStatus and go straight to autoAuthentication. Handle V19 plaintext packets alongside V2/V3 encrypted packets in parseEncPackets. Set packet version to V2 to match device protocol.

PowerStream now decodes inverter_heartbeat protobuf messages from V13 BLE frames into 16 sensor fields (PV1/PV2 power/voltage/current/temp, battery level/power/temp, grid power/voltage/current/frequency, inverter temp).

Key fixes to the BLE protocol layer:
- All frame versions (V2, V3, V13) are now AES-CBC decrypted uniformly; V13 was previously treated as plaintext, causing parse failures and 5-error disconnects
- Only cmdId=0x01 heartbeats are parsed; cmdId=0x04 is a different message type that produces garbage when decoded as inverter_heartbeat

The wn511_sys_pb2.py was generated with protoc 33.2 which adds a runtime_version.ValidateProtobufRuntimeVersion check requiring protobuf >= 6.x. This breaks the import chain in HA environments with older protobuf, causing "Config flow could not be loaded".

Regenerated with protoc 24.4 to match all other _pb2.py files.

Heartbeat cmdId is 0x01 (110 fields, correct values), not 0x04 (60 fields, garbage values). Also fix auth flow to send getAuthStatus before autoAuthentication, use response=False for GATT writes to avoid 20s hang, subscribe to notifications once in initBleSessionKey, and handle max_attempts=0 correctly for BLE-level retries.

The PowerStream sends two BLE heartbeat types: cmdId=0x01 (inverter_heartbeat with full PV/battery/grid telemetry) and cmdId=0x04 (inv_heartbeat_type2 with settings/status and nested battery module data). Previously only cmdId=0x01 was handled.

Added inv_heartbeat_type2 and psdr_heartbeat_type3 proto definitions from the decompiled EcoFlow app protos, with [packed=true] for proto3 wire compatibility. Added throttled reply (10s) on cmdId=0x04 to keep the device sending data (DPU pattern). Added periodic heartbeat request loop (30s) to trigger cmdId=0x01 after auth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,33 @@ first charging task</sup>
 </details>
 
 <details>
+<summary><b>PowerStream</b></summary>
+
+<br>
+
+| *Sensors*            | *Sliders*        | *Selects*             |
+|----------------------|------------------|-----------------------|
+| Battery Level        | Load Power       | Power Supply Priority |
+| Battery Power        | Charge Limit     |                       |
+| Battery Temperature  | Discharge Limit  |                       |
+| PV (1) Power         |                  |                       |
+| PV (1) Voltage       |                  |                       |
+| PV (1) Current       |                  |                       |
+| PV (1) Temperature   |                  |                       |
+| PV (2) Power         |                  |                       |
+| PV (2) Voltage       |                  |                       |
+| PV (2) Current       |                  |                       |
+| PV (2) Temperature   |                  |                       |
+| Inverter Power       |                  |                       |
+| Inverter Voltage     |                  |                       |
+| Inverter Current     |                  |                       |
+| Inverter Frequency   |                  |                       |
+| Inverter Temperature |                  |                       |
+| LLC Temperature      |                  |                       |
+
+</details>
+
+<details>
 <summary><b>Wave 3</b></summary>
 
 <br>
@@ -327,6 +354,8 @@ first charging task</sup>
 | Cell Temperature (disabled)      |                 |            |
 | Fan Level                        |                 |            |
 | Drainage Mode                    |                 |            |
+| In Drainage                      |                 |            |
+| Display Temperature Type         |                 |            |
 | Input Power                      |                 |            |
 | Output Power                     |                 |            |
 | Indoor Supply Air Temperature    |                 |            |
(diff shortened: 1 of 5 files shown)


GnoX:
Not only it works, but it seems that it fixes problems with SHP2 switches - we were having a discussion about it here https://github.com/GnoX/ha-ef-ble/pull/54#issuecomment-3941360169 - after SHP2 received command, it always sent a packet with invalid header length causing previous implementation to skip a lot of messages while buffering for the whole invalid length.

Btw, the description you copied from the original MR is mostly incorrect (AI generated) and few statements are no longer valid for this MR as I removed messages,  reverted some of the changes as they were causing issues for other devices, reextracted protobufs and identified what was causing parsing errors.

PR #195, opened 2026-02-28
Use serial number for entity unique id instead of name due to possible conflicts

### LINKED ISSUE #194: [Bug]: adding second river 3 plus - no sensors shown
### What happened?

I have added a second river 3 plus box today (it was detected automatically) and I managed to add it using my ecoflow id. Once added it reports 5 entities and when I click to see them, nothing appears. No sensors or controls are displayed either.

However the first ecoflow river 3 which I added two weeks back is visible with all its 29 controls/entities shown, but the second one does not show any sensors.

Could you please let me know what the issue could be please? Is there an issue where we cannot manage two identical devices?

0.6.7

Ecoflow river 3 plus (second device)

HA running on Intel mini pc

Internal bluetooth adapter

Yes (Please write down the changed settings in the next step)

```text
No changes made only added second device which was auto detected
```

### Relevant log output

```shell

```

GnoX:
This PR changes the way we determine unique ids for all entities to serial numbers. Names are not greate because they can cause collisions as shown in #194. 

Resolves #194

### DIFF
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -6,13 +6,14 @@ from functools import partial
 import homeassistant.helpers.issue_registry as ir
 from homeassistant.components import bluetooth
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import CONF_ADDRESS, CONF_TYPE, Platform
+from homeassistant.const import CONF_ADDRESS, Platform
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import (
     ConfigEntryError,
     ConfigEntryNotReady,
 )
-from homeassistant.helpers.device_registry import DeviceInfo
+from homeassistant.helpers import device_registry as dr
+from homeassistant.helpers import entity_registry as er
 
 from . import eflib
 from .config_flow import CONF_COLLECT_PACKETS, ConfLogOptions, LogOptions, PacketVersion
@@ -24,7 +25,6 @@ from .const import (
     DEFAULT_CONNECTION_TIMEOUT,
     DEFAULT_UPDATE_PERIOD,
     DOMAIN,
-    MANUFACTURER,
 )
 from .eflib.connection import (
     AuthFailedError,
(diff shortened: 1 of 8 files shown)


rabits (commented):
Hmmm, this is a tricky one... This changes how devices making sensor names and I'm not sure SN is a good primary sensor name instead of the device name for all the devices. Surely name of the device could be identical - but the users usually see name, not the SN...

Maybe instead it's better to modify the specific river 3 device or allow user to pick the right name? Or use meaningful (last 4 symbols) part of SN?

Not sure about this one overall... Maybe using SN is a good idea, just worrying it will seriously change the existing interface.

GnoX:
> This changes how devices making sensor names

This is not true - `unique_id` is not visible to the user, `entity_id` is. If `entity_id` is not overriden then it uses device name as a prefix and translation as suffix - this is how sensors are named right now. This mean that no entity names will change as a result of this, it only changes how they are stored in registry.

rabits:
> This is not true - `unique_id` is not visible to the user, `entity_id` is.

Yeah maybe that's my misunderstanding. So what's used in the automation? Automation surely has to use unique id to refer to a specific sensor of the device. So I don't get what is the real difference for the user between unique_id and entity_id - when we changing unique id - prepared automation will still be under a risk to fail (I suppose there is no such heuristic in HA to migrate the automation as well).

GnoX:
`unique_id` is only used as a unique identifier that is only used to identify the entity for storing in the entity registry, you cannot even use this in automations. These should be unique for device-sensor combinations, which weren't before this change. User cannot do anything about this which makes it impossible to register some devices that have the same serial number suffixes (as in #194). What user can override is that `entity_id` and this one can be duplicate - it will not fail if you register two entities with the same id (one you register later will automatically get postfixed with its index, i.e. `_2`.

You can look at `config/.storage/core.entity_registry` and see that there's `id`, `unique_id` and `entity_id` in there.

Config entries create their own config entry unique `id` - this is the thing you use for automations using Device selectors. This is unique per every entity you create and is different each time you register the device in this integration whereas `unique_id` should be the same if you remove and add it again.

Constructed entity ids (what you would use for state selectors, i.e. `sensor.ef_<device_name>_<sensor>`) are determined by that entity_id - we are still using defaults, we never touched these.

I'm using my fork where I already released it on my main HA instance and everything is still working.

rabits:
So why unique_id is not getting this _2 suffix automatically as entity_id? Maybe I'm missing something?

GnoX:
Because that's what unique id is for - it is so that you cannot register the same entity for a device twice if you for example had two instances of the same integration. Entity id is configurable by user so there is a chance user could name it the same way as integration default so it deduplicates those and that's why entity_id cannot be used to bind an entity to a device. Same for config entry `id` - it only uniquely identifies a config, not a device entity.

rabits:
Ok, so if I understand correctly of what you saying - that entity_id will not change at all, just the unique id will have those SN modifications? If we will add new device to the roster - entity_id will have the same name-based value? Then I suppose it's fine...

GnoX:
Yep, everything will stay the same, only the unique identifier will get migrated to have sn in there.

Btw, heres excerpt about unique_id from https://developers.home-assistant.io/docs/core/entity/?_highlight=unique_id#registry-properties
> A unique identifier for this entity. It must be unique within a platform (like light.hue). It should not be configurable or changeable by the user

And a link that goes to https://developers.home-assistant.io/docs/entity_registry_index#unique-id-requirements that basically says that we were doing it wrong before this.

rabits (approved):
Yep, fixed my lack of understanding - now it's clearly a great change!)

PR #201, opened 2026-03-03
Remove conversion to hex for unused logging operations causing unnecessary cpu usage

### LINKED ISSUE #199: CPU Usage for disabled logger in custom_components/ef_ble/eflib/connection.py
Hi!

I may have found something that might reduce the cpu usage:

in custom_components/ef_ble/eflib/connection.py there is this:

```
            self._logger.log_filtered(
                LogOptions.ENCRYPTED_PAYLOADS,
                "parseSimple: Data: %r",
                bytearray(data).hex(),
            )
```
it seems that bytearray(data).hex() is always executed even if no log option is given.

i did a check for the loggings:
```
        if getattr(self._logger, "is_enabled", lambda *_: False)(LogOptions.ENCRYPTED_PAYLOADS):
            self._logger.log_filtered(
                LogOptions.ENCRYPTED_PAYLOADS,
                "parseSimple: Data: %r",
                bytearray(data).hex(),
            )
```

CPU usage in the affected code path was reduced significantly based on callgrind profiling:

Before optimization:
parseEncPackets + checksum + _process_byte together accounted for ~32.535% of total CPU time.

After optimization:
The same functions account for only ~0.839% of total CPU time.

This corresponds to an approximate 97.4% reduction in CPU consumption for this code path.
Hint: it was just a quick compare - in my n150 with pl1 on 6W i see about ~0.5% deeper CPU.

Maybe you could give this a look?

Thank you kind regards

GnoX:
This PR removes byte conversion to hex strings for logging calls that are not enabled - string is only materialized if it is actually printed out.

Resolves #199

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -577,7 +577,7 @@ class Connection:
         self._logger.log_filtered(
             LogOptions.ENCRYPTED_PAYLOADS,
             "parseSimple: Data: %r",
-            bytearray(data).hex(),
+            data,
         )
 
         header = data[0:6]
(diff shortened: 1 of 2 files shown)


rabits (approved):
Great optimization! Thank you

PR #206, opened 2026-03-08
Add support for Delta Pro

### LINKED ISSUE #139: Delta Pro (V1) support
Dear Developers Team,
Could you please consider to add Delta Pro (V1) support?
The last update (FW: V1.0.1.144) allows us to discover BLE connection of the device, but it is "Failed to set up".
Thank you!

GnoX:
This PR adds suport for Delta Pro requested and tested in #139 and support for nordic uart comm protocol used by some first gen devices.

Resolves #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,34 @@ first charging task</sup>
 </details>
 
 <details>
+<summary><b>Delta Pro</b></summary>
+
+<br>
+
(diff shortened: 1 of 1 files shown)


rabits (approved):
Great! Can't test effectively, but seems good)

PR #212, opened 2026-03-10
Adds individual circuit on/off switches for SHP2

ecammit:
This is the culmination of a bunch of work and testing done with @GnoX to add individual circuit on/off switches.

Additional changes:
- Changes device disconnect logic to reset the failed packet counter if a valid packet is received and parsed successfully. The SHP2 periodically sends truncated packets that causes frequent device reconnects without this.
- Handles split-phase circuits (see below).

Importantly, this ensures that split phase circuits (which is when two separate 120v circuits are linked together to form a 240v circuit fed by a 2-pole breaker) are correctly enabled/disabled in-tandem. Otherwise if the individual circuits are linked together for split-phase and are not in the same state (one is off and the other is on), the SHP2 will periodically (every few seconds) toggle the circuit to the opposite state and back to the original state again within a couple milliseconds. This "hiccup" results in excessive relay use and causes continual brown-out conditions in the circuit. The code in this PR avoids this by:
 - being aware of the split-phase circuits
 - sending a single command to switch both circuit states in a split-phase pair

Superceeds PRs #54 and #56.

### DIFF
--- a/custom_components/ef_ble/switch.py
+++ b/custom_components/ef_ble/switch.py
@@ -163,10 +163,10 @@ async def async_setup_entry(
         EcoflowSwitchEntity(device, switch_desc)
         for switch_desc in SWITCH_TYPES
         if (
-            isinstance(switch_desc, EcoflowSwitchEntityDescription)
-            or (
-                hasattr(device, switch_desc.key)
-                and hasattr(device, f"enable_{switch_desc.key}")
+            hasattr(device, switch_desc.key)
+            and (
+                isinstance(switch_desc, EcoflowSwitchEntityDescription)
+                or hasattr(device, f"enable_{switch_desc.key}")
             )
         )
     ]


custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -483,16 +549,63 @@ async def set_circuit_power(self, circuit_id, enable):
...
+            and getattr(self, f"circuit_{circuit_id}_split_link", None) is not None
+            and (
+                getattr(self, f"circuit_{circuit_id}_is_split") is False
+                or getattr(self, f"circuit_{circuit_id}_split_link") > circuit_id
+            )
+        )
+
+    def get_circuit_switch_additional_properties(self, circuit_id: int):
+        """Get the list of additional device properties to listen for state updates for the circuit switch (e.g. split circuit info)"""
+        return [f"circuit_{circuit_id}_is_split", f"circuit_{circuit_id}_split_link"]
```
GnoX: Is this used anywhere?

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -483,16 +549,63 @@ async def set_circuit_power(self, circuit_id, enable):
...
+
         payload = ppas.SerializeToString()
         packet = Packet(0x21, 0x0B, 0x0C, 0x21, payload, 0x01, 0x01, 0x13)
 
         await self._conn.sendPacket(packet)
+
+    def is_circuit_switch_available(self, circuit_id: int):
+        """
+        Circuit switch is available if the device is connected, the state is available, and the circuit is not a split circuit
+           or it's the main circuit of a split circuit (circuit with higher circuit id is the sub circuit)"""
```
GnoX:
```suggestion
        """
        Circuit switch is available if the device is connected, the state is available, and the circuit is not a split circuit
        or it's the main circuit of a split circuit (circuit with higher circuit id is the sub circuit)
        """
```

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -483,16 +549,63 @@ async def set_circuit_power(self, circuit_id, enable):
...
+        return (
+            self.is_connected
+            and getattr(self, f"circuit_{circuit_id}", None) is not None
+            and getattr(self, f"circuit_{circuit_id}_is_split", None) is not None
+            and getattr(self, f"circuit_{circuit_id}_split_link", None) is not None
+            and (
+                getattr(self, f"circuit_{circuit_id}_is_split") is False
+                or getattr(self, f"circuit_{circuit_id}_split_link") > circuit_id
+            )
+        )
```
GnoX: This is a bit hard to read, maybe take those getattrs into variables and reuse them?

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -467,6 +522,17 @@ async def data_parse(self, packet: Packet) -> bool:
...
+    def _on_connection_state_change(self, state: ConnectionState):
+        if state == ConnectionState.AUTHENTICATED:
+
+            async def _update_controls():
+                for field in self._fields:
+                    value = getattr(self, field.public_name)
+                    if value is not None:
+                        self.update_state(field.public_name, value)
+
+            self._conn.call_later(_update_controls(), delay=10)
```
GnoX: I put this here mostly for testing, better way to do this would be to do it in connection as I did experience a similar issue with my devices. Better way be to buffer updated properties for few seconds longer in update callbacks.

custom_components/ef_ble/switch.py
```diff
@@ -177,7 +210,15 @@ def state_updated(self, state: bool | None):
 
     @property
     def available(self):
-        return self._device.is_connected and self._on_off_state is not None
+        return (
+            self._device.is_connected
+            and self._on_off_state is not None
+            and (
+                not isinstance(self.entity_description, EcoflowSwitchEntityDescription)
+                or not getattr(self.entity_description, "available", None) is not None
```
GnoX: Now this is very confusing, I think this was meant to be

```suggestion
                or getattr(self.entity_description, "available", None) is None
```

GnoX: Also, entity_description is invariant so we don't need to check those two conditions every time avability is called, either make it a variable or take it out as @cached_property

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -483,16 +549,63 @@ async def set_circuit_power(self, circuit_id, enable):
...
+            self._logger.warning(
+                "%s: Cannot set circuit power for circuit %d because split circuit info is not available",
+                self._address,
+                circuit_id,
+            )
+            return
+        if is_split and (split_link < 1 or split_link > self.NUM_OF_CIRCUITS):
+            self._logger.warning(
+                "%s: Cannot set circuit power for circuit %d because split link circuit id %d is invalid",
+                self._address,
```
GnoX: Isn't the mac address redundant? Device logger should already be printing it.

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -32,6 +33,12 @@ class ControlStatus(IntFieldValue):
...
     UNKNOWN = -1
 
+
+class CircuitState(IntFieldValue):
+    """Circuit state enum (0=OFF, 1=ON)"""
+
+    UNKNOWN = -1
+
     OFF = 0
     ON = 1
```
GnoX, line 36: Something is not right here. I think you accidentally removed enum options for ForceChargeStatus?

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -483,16 +549,63 @@ async def set_circuit_power(self, circuit_id, enable):
...
+        Circuit switch is available if the device is connected, the state is available, and the circuit is not a split circuit
+           or it's the main circuit of a split circuit (circuit with higher circuit id is the sub circuit)"""
+        return (
+            self.is_connected
+            and getattr(self, f"circuit_{circuit_id}", None) is not None
+            and getattr(self, f"circuit_{circuit_id}_is_split", None) is not None
+            and getattr(self, f"circuit_{circuit_id}_split_link", None) is not None
+            and (
+                getattr(self, f"circuit_{circuit_id}_is_split") is False
+                or getattr(self, f"circuit_{circuit_id}_split_link") > circuit_id
```
GnoX: Question - does this need to go unavailable now that it switches both circuits for any of the linked switches?

GnoX:
BTW, I intended to make this PR myself and clean up some things over this weekend so you got ahead of me here. I didn't point out some things in your PR into my dev branch as it figured it would be quicker to do these myself.

ecammit:
> BTW, I intended to make this PR myself and clean up some things over this weekend so you got ahead of me here. I didn't point out some things in your PR into my dev branch as it figured it would be quicker to do these myself.

No worries. I was just working on adding some more SHP2 controls and was having trouble navigating the divergence between your branch and the official, so I decided to open a PR here to prevent my own code rot.

custom_components/ef_ble/eflib/devices/shp2.py
ecammit: > Question - does this need to go unavailable now that it switches both circuits for any of the linked switches?

Technically no. You'd just have two duplicate switches for essentially the same circuit. They'll both trigger together. I just thought that keeping one unavailable might be less confusing as users might realize that it's unavailable because it's part of a linked circuit.

My ideal state would be to disable the entity for one of the circuits entirely so it doesn't show up, as I've manually done for my own setup. However, the current way entities are managed doesn't allow for delaying entity addition until after the device is queried.

Your choice on how to handle this. I could definitely keep both switches available.

GnoX:
Yeah, don't worry, I didn't forget about it, I was just waiting for those dependent PRs for other devices to be reviewed and merged. 

I'm also cooking up something to make adding controls and indexed sensors easier which I meant to do ages ago, but device requests just kept coming :) I will finally have some time to do something about it this weekend.

custom_components/ef_ble/eflib/devices/shp2.py
GnoX: > My ideal state would be to disable the entity for one of the circuits entirely so it doesn't show up, as I've manually done for my own setup. However, the current way entities are managed doesn't allow for delaying entity addition until after the device is queried.

Well, the problem with this is handling what happens if you reconfigure those links. You can't easily add/remove entities dynamically. 
I think what I would do is display each circuit as a connected device and let user be able to choose which circuits he wants to see from the reconfigure config flow - similar to how we implemented extra batteries.

ecammit: I'll remove it for now so both circuits are available. It'll simplify the code and be easier to maintain.

custom_components/ef_ble/eflib/devices/shp2.py
ecammit: Correct. I was just blindly following what the existing debug() lines did without checking. I'll update the debug lines to remove the redundancy also.

custom_components/ef_ble/eflib/devices/shp2.py
ecammit: I tested and it works without this, so I just removed it.

custom_components/ef_ble/eflib/devices/shp2.py
ecammit: Fixed

custom_components/ef_ble/eflib/devices/shp2.py
ecammit: Fixed

ecammit:
Note: I finally tracked down the gremlin that would make the circuits sometimes appear in an off state immediately after turning the circuit on. It was the `TransformIfMissing` wrapping the `transform` methods on the `pb_field()`. The issue is that the panel reports back a message with a bare `load_ch_switch_cause` attribute inside the `ch##_sta` that is expected to contain the `load_sta` attribute that indicates the circuit status. The timing of the messages is sometimes such you'll receive `load_sta` before `load_ch_switch_cause` or you may receive both in the same message.

So it was fine when you'd get a reply of:

```
load_incre_info {
  hall1_incre_info {
    ch10_info {
      load_rly_cnt: 76
    }
    ch12_info {
      load_rly_cnt: 78
    }
    ch10_sta {
      load_sta: LOAD_CH_POWER_ON
      load_ch_switch_cause: 1
    }
    ch12_sta {
      load_sta: LOAD_CH_POWER_ON
      load_ch_switch_cause: 1
    }
  }
}
```

But if you got two separate messages:

```
load_incre_info {
  hall1_incre_info {
    ch10_info {
      load_rly_cnt: 76
    }
    ch12_info {
      load_rly_cnt: 78
    }
    ch10_sta {
      load_sta: LOAD_CH_POWER_ON
    }
    ch12_sta {
      load_sta: LOAD_CH_POWER_ON
    }
  }
}
```

...that is imediately followed by:


```
load_incre_info {
  hall1_incre_info {
    ch10_info {
      load_rly_cnt: 77
    }
    ch12_info {
      load_rly_cnt: 79
    }
    ch10_sta {
      load_ch_switch_cause: 0
    }
    ch12_sta {
      load_ch_switch_cause: 0
    }
  }
}
```

... the previous `TransformIfMissing` was showing the circuit as `off` after processing the `load_ch_switch_cause` message.

GnoX:
> Note: I finally tracked down the gremlin that would make the circuits sometimes appear in an off state immediately after turning the circuit on. It was the TransformIfMissing wrapping the transform methods on the pb_field(). The issue is that the panel reports back a message with a bare load_ch_switch_cause attribute inside the ch##_sta that is expected to contain the load_sta attribute that indicates the circuit status. The timing of the messages is sometimes such you'll receive load_sta before load_ch_switch_cause or you may receive both in the same message.

Yeah, that was the thing I fixed when I mentioned it in https://github.com/GnoX/ha-ef-ble/discussions/59#discussioncomment-15927126 - it is fixed in my branch.

rabits:
Very nice! That's what I was unable to wrap my head around - those split-phases awareness! Thank you for making that happen)

ecammit:
I made a few cleanup changes to mirror the select's existing `availability_prop` and `set_state` functionality in switch. I also added the translations entries that I missed the first time around.

custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -445,54 +508,72 @@ async def data_parse(self, packet: Packet) -> bool:
...
         )
         sta.load_sta = (
             pd303_pb2.LOAD_CH_POWER_ON if enable else pd303_pb2.LOAD_CH_POWER_OFF
         )
         sta.ctrl_mode = pd303_pb2.RLY_HAND_CTRL_MODE
-        payload = ppas.SerializeToString()
-        packet = Packet(0x21, 0x0B, 0x0C, 0x21, payload, 0x01, 0x01, 0x13)
 
-        await self._conn.sendPacket(packet)
+        # If it's a split circuit, also set the linked circuit to the same state because they will be turned on/off together
```
GnoX, line 568: This line is quite a bit long, I forgot to put max line length of 79 to pyproject, but can take care of this, it is a nitpick.

custom_components/ef_ble/switch.py
```diff
@@ -124,6 +135,20 @@
         key="emergency_reverse_charging",
         name="Emergency Reverse Charging",
     ),
+    # SHP2 Circuit switches
+    *[
+        EcoflowSwitchEntityDescription[shp2.Device](
+            key=f"circuit_{i}",
+            name=f"Circuit {i}",
+            translation_key="circuit_is_enabled",
```
GnoX, line 143: I think that translation key won't be used if you have name set, right?

ecammit: Other way around. If the `translation_key` is present and there is a translation provided for the current locale, the translation is used. Otherwise it falls back to the `name` provided. I just verified this locally.

ecammit:
> Looks good, thanks.
> 
> I hope it won't be a nightmare merging this back into my fork :)

I have some more controls for SHP2 that are based off this branch, that will be in a separate PR shortly after this, so it's your choice if you want to merge this now or later.

rabits:
Let me check it tomorrow on my secondary SHP2 and after that we can merge.

rabits:
Hmmm, I just checked and looks like the circuits are unavailable... Nothing much suspicious in the log, maybe my firmware is not good enough? But last changes I checked - at least showed those switches enabled, so not quite sure what's up here...
```
...
2026-03-12 19:52:36.071 INFO (MainThread) [homeassistant.components.switch] Setting up ef_ble.switch
2026-03-12 19:52:36.072 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_1
2026-03-12 19:52:36.074 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_2
2026-03-12 19:52:36.077 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_3
2026-03-12 19:52:36.078 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_4
2026-03-12 19:52:36.080 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_5
2026-03-12 19:52:36.082 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_6
2026-03-12 19:52:36.083 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_7
2026-03-12 19:52:36.084 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_8
2026-03-12 19:52:36.086 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_9
2026-03-12 19:52:36.089 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_10
2026-03-12 19:52:36.090 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_11
2026-03-12 19:52:36.092 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new switch.ef_ble entity: switch.ef_hd30377_circuit_12
...
```
I can provide more details if needed.

rabits:
Actually I've checked in an hour - and they've become available on one of the panels (second one is still shows unavailable state for all the switches)... So probably the required package with the state is not getting through, although I tried to disable the throttling (set update time to 0).

rabits:
But despite this issue with "unavailable" state in the beginning - the switches are working fine. Maybe instead of waiting for the state we can actively request the status during initialization somehow?

GnoX:
Yeah, that's one problem we were discussing with @ecammit - device callbacks get triggered before they are registered in sensor code which means that they miss first few updates which is bad for incremental messages.

ecammit:
Generally speaking, I've gotten them available within the first 10-15s.

I did find that the integration was very much dependent upon the proximity of the bluetooth controller to the panel or the BLE messages can get truncated. I have an ESPHome bluetooth proxy that's located pretty far away through several walls and for some reason that gets chosen sometimes to connect to the SHP2 over the direct USB dongle that's less than 1m away with clear line-of-sight to the panel's antenna.

rabits:
Maybe we can maintain the internal cache of received data? Effectively we already do for the known sensors, but maybe we can have the same for unknown ones, so the state could be received after the fact?

GnoX:
Well, we can could get the last known value from HA DB, but that could be out of sync by the time it comes back online.
Let's first test if, instead of the initial value being None in the switch entity, we assign it to the value of registered device property. This way we don't have to wait for the callback, but have the latest state of the device as the initial value of the switch. Not sure why we didn't do that from the start. Will you take care of this @ecammit?

ecammit:
> Well, we can could get the last known value from HA DB, but that could be out of sync by the time it comes back online. Let's first test if, instead of the initial value being None in the switch entity, we assign it to the value of registered device property. This way we don't have to wait for the callback, but have the latest state of the device as the initial value of the switch. Not sure why we didn't do that from the start. Will you take care of this @ecammit?

Is the issue /only/ that the state is unavailable or is it also that the circuit cannot be toggled right away?

I think part of this is that we disabled the switch using the availability property, which shows the control as `unavailable` regardless of whether it has a known on/off state already. Because I think we know the state of the circuit from the device much sooner than we know the split-phase config. I'll have to see if there is another way to disable an HA control programmatically outside of setting availability so that it at least shows the current state, but disables the switch from being toggled until we have the latest split circuit config. I think this might be possible.

My concern with using last state for the split-phase info is that any automation (or even a human) could toggle a circuit before we refresh from the latest circuit config. This has the possibility of putting the panel into the state we observed where the circuit relay continually fast-cycles. It shortens the relay life of the circuit and will damage connected equipment from continual "brown-outs".

We could potentially detect this bad state if we monitor the relay counts for each circuit and if the count is increasing without the state of the circuit changing, we can issue a corrective set command for both linked circuits to set it to the current circuit state and it should resolve the problem without anyone knowing. I'll probably look into adding this in a future PR.

Additionally, if we still want to use the cached state of the circuits at startup, maybe we could add a time constraint so that we don't re-use data from some timeframe like older than 5 minutes ago. This will help with the circuits becoming available after restarts/reconnects, while protecting against assuming stale data for cases where the integration was offline or disabled for an extended period.

ecammit:
There is an "assumed_state" property in HA for entities that I think should be used for exactly the case of re-using the prior state.

GnoX:
> I think part of this is that we disabled the switch using the availability property, which shows the control as unavailable regardless of whether it has a known on/off state already. Because I think we know the state of the circuit from the device much sooner than we know the split-phase config. I'll have to see if there is another way to disable an HA control programmatically outside of setting availability so that it at least shows the current state, but disables the switch from being toggled until we have the latest split circuit config. I think this might be possible.

Well, what we can do is just return early in the enable function - this will switch the state in gui but returns back in few seconds without doing anything.

I wouldn't do anything with cached state just yet, but we should at least initialize entities from the device state. If they're set, it's safe, if it's not, it's None anyways.

ecammit:
@rabits what is your bluetooth setup? Are you using a bluetooth proxy or a direct bluetooth controller? How close is your panel to the bluetooth controller?

I ask because the only way I can seem to replicate the circuits being unavailable for any length of time is if I disable the bluetooth controller near the panel and use a ESP bluetooth proxy that's far away. Otherwise the circuits become available almost immediately (~1-2s after the other sensors).

I'm just concerned that loading cached info is potentially just masking a bluetooth connection issue that will only frustrate users when the integration appears to sometimes work, but other times just silently drop things. If the circuits were not consistently showing up as available (with proper state) immediately after reload on my setup, I'd 100% agree that it's some timing issue where cached values may help. So I'm just curious what your setup is.

I'll do some further testing to see if it is using a ESP bluetooth proxy or simply the low RSSI due to distance and obstruction, but based upon the past logging, I think the panel is sending out the circuit config and it's getting truncated and jumbled with other messages by the time it gets to the integration code. This could also be MTU issues (that I know GnoX was looking into). But the panel doesn't seem to resend the circuit info again when it fails to receive any ACK back. So if the circuit info packets are dropped as invalid, the circuits just sit as unavailable until something eventually triggers the info to get sent again.

ecammit:
Update: I just disabled all bluetooth controllers except for an ESP32 running bluetooth proxy right next to the panel. The sensors are slow to become available after a reload and the circuits remain unavailable... which likely means the packets containing the circuit config info were dropped.

So there's definitely some low-level issue with ESP bluetooth proxies.

ecammit:
Screen capture of load times using the USB bluetooth controller (no ESP bluetooth proxy). Note, this is using my current #217 PR that includes this code.

https://github.com/user-attachments/assets/13c560d4-94cb-46a1-b713-72b8b2df5438

Circuits are:
- Air Conditioner [off]
- Dryer
- Kitchen Inner Wall
- Kitchen Outer Wall
- Subpanel
- Water Heater

The all are split-phase except the kitchen ones. I manually disabled the duplicate switches for the split-phase, so only one appears.

rabits:
Yeah my setup is ESP ble proxies nearby the panels. One proxy for each panel. Just exactly as described here: I suppose what you saw with ble proxies - is what I see as well. Do you think you can identify what kind of packets are dropped on proxies? I think quite a bit of folks are using ble proxy-only setup - but in general functionality is working somehow, so if you're ok with the state for now - please just mark the issue in README and create a ticket, so we would not loss it - and I will be able to merge it to unblock the next PR's.

ecammit:
I did just add the code to default the select and switch values to the current device property value, with a fallback to None. Ultimately it didn't help the ESP performance, but @GnoX is correct that they should probably be set to starting values this way.

> please just mark the issue in README and create a ticket, so we would not loss it 

I'll open an issue and add another commit with a README link. Stay tuned.

Now that I know BT proxy is an issue, I should hopefully have all I need to actually track it down, given I have reference hardware of it working as well.

GnoX:
> I did just add the code to default the select and switch values to the current device property value, with a fallback to None. Ultimately it didn't help the ESP performance, but @GnoX is correct that they should probably be set to starting values this way.

Yeah, didn't think it would improve it, but at least the initial state is little bit more consistent, thanks.

I would like to help further with this, but while I have a good variety of ESP boards, none of my EF devices are incremental or exhibit similar behavior so I have no way to reproduce this.

ecammit:
Okay, issue opened and linked in README under the SHP2 section.

ecammit:
Oof. Just discovered all the controls are appearing for newly created devices that are not an SHP2 (ex. DPU).

custom_components/ef_ble/switch.py
```diff
@@ -137,8 +162,13 @@ async def async_setup_entry(
...
         for switch_desc in SWITCH_TYPES
-        if hasattr(device, switch_desc.key)
-        and hasattr(device, f"enable_{switch_desc.key}")
+        if (
+            isinstance(switch_desc, EcoflowSwitchEntityDescription)
+            or (
+                hasattr(device, switch_desc.key)
+                and hasattr(device, f"enable_{switch_desc.key}")
+            )
+        )
```
GnoX, line 171: Yeah, I missed this - should be
```suggestion
        if (
            hasattr(device, switch_desc.key)
            and (
                isinstance(switch_desc, EcoflowSwitchEntityDescription)
                or hasattr(device, f"enable_{switch_desc.key}")
            )
        )
```

ecammit: Yeah, that fixed it. I was definitely reading that wrong before. Thanks for finding it!

ecammit:
I think it's probably all set now.

PR #214, opened 2026-03-10
Add support for Delta 3 Ultra Plus

GnoX:
Added Delta 3 Ultra Plus device


rabits:
Sorry, could I ask for description in PR to keep the history and make sure changes are what you want to do here?

GnoX:
Well, it just adds support for Delta 3 Ultra Plus which is already in the title so didn't want to repeat it for no reason. Other changes are just internal function renames, nothing else.

Issue #87, opened 2025-10-18
Controls/switches for SHP2


MatthewJB2022:
Is it possible to get more than just sensor data for SHP2? Would love to be able to toggle operation mode and battery charging settings

rabits:
Hi @MatthewJB2022 , yeah I started to work on controls in #85 , but right now I see an issue with optimizations we have (or maybe incorrect protocol who knows). In general controls are working, but their initial state is not always correct...

ecammit:
I think this can be closed now that #217 is merged and is in v0.7.1.

PR #85, opened 2025-10-12
Added control to change backup reserve on SHP2

rabits:
This change adds a few sensors and a way to control backup reserve to SHP2 and channel enable/disable. Unfortunately controls are still not that stable, because I think alot of important messages is skipped due to the throttle optimization, but still need to figure out why is that...

Although it's not perfect - I think it's better then nothing. The switches works (not every time, but at least allows to control state remotely). So I think it's good enough to be merged.

PR #217, opened 2026-03-12
Several new SHP2 controls

ecammit:
Added shp2 items:
    - Channel enable/disable
    - Channel force charge
    - EPS mode
    - Smart Backup Mode
    - Backup Reserve Level
    - Max Charge Level
    - Power Type
    - Storm Mode
    - Grid Status
    - AC Charge Speed (Watts)

Also added step_value_prop to number sliders to allow it to change per device

NOTE: This is built off of PR #212. Until that is merged, this PR may be a little more difficult to review.

Closes Issue: #87 
Incorporates PR: #212 
Superceeds PR: #85

### DIFF
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -33,7 +33,6 @@ class _BinarySensorKwargs(TypedDict, total=False):
     translation_placeholders: dict[str, str]
     indexed_range: range
     entity_category: EntityCategory
-    icon: str
 
 
 def _make_desc(
@@ -142,8 +141,8 @@ _BINARY_SENSORS: Final[dict[str, BinarySensorEntityDescription]] = {
         connectivity, "channel_hw_connected", enabled=False
     ),
     # SHP2 generic binary sensors
-    "grid_status": connectivity("grid_status", enabled=True, icon="mdi:connection"),
-    "storm_mode": safety("storm_mode", enabled=True, icon="mdi:flash-alert"),
+    "grid_status": connectivity("grid_status", enabled=True),
+    "storm_mode": safety("storm_mode", enabled=True),
 }
 
 BINARY_SENSOR_TYPES: Final[dict[str, BinarySensorEntityDescription]] = (
--- a/custom_components/ef_ble/eflib/devices/shp2.py
+++ b/custom_components/ef_ble/eflib/devices/shp2.py
@@ -666,14 +666,18 @@ class Device(DeviceBase, ProtobufProps):
     async def set_eps_mode(self, value: bool):
         self._logger.debug("set_eps_mode: %d", value)
 
+        if value and self.smart_backup_mode != SmartBackupMode.NONE:
+            # App forces setting of operating mode to NONE when EPS is enabled.
+            # We set this to NONE first or the SHP2 will sometimes report a grid outage if
+            # we set it all in one PPS command.
+            # Note: unlike operating mode force charge is allowed with EPS mode
+            ppas_sbm = pd303_pb2.ProtoPushAndSet()
+            ppas_sbm.smart_backup_mode = SmartBackupMode.NONE
+            await self._send_config_packet(ppas_sbm)
+
         ppas = pd303_pb2.ProtoPushAndSet()
 
         ppas.eps_mode_info = value
-        if value:
-            # App sets operating mode to NONE when EPS is enabled
-            # However, unlike operating mode force charge is allowed with EPS mode
-            ppas.smart_backup_mode = SmartBackupMode.NONE
-
         await self._send_config_packet(ppas)
         return True
 
--- a/custom_components/ef_ble/number.py
+++ b/custom_components/ef_ble/number.py
@@ -304,8 +304,6 @@ NUMBER_TYPES: list[EcoflowNumberEntityDescription] = [
     ),
     EcoflowNumberEntityDescription[shp2.Device](
         key="backup_reserve_level",
-        name="Backup Reserve",
-        icon="mdi:battery-arrow-down-outline",
         device_class=NumberDeviceClass.BATTERY,
         native_unit_of_measurement=PERCENTAGE,
         native_step=1.0,
@@ -318,8 +316,6 @@ NUMBER_TYPES: list[EcoflowNumberEntityDescription] = [
     ),
     EcoflowNumberEntityDescription[shp2.Device](
         key="backup_charge_limit",
-        name="Charge Limit",
-        icon="mdi:battery-arrow-up",
         device_class=NumberDeviceClass.BATTERY,
         native_unit_of_measurement=PERCENTAGE,
         native_step=1.0,
--- a/custom_components/ef_ble/sensor.py
+++ b/custom_components/ef_ble/sensor.py
@@ -58,7 +58,6 @@ class _SensorKwargs(TypedDict, total=False):
     indexed_range: range
     entity_category: EntityCategory
     state_attribute_fields: list[str]
-    icon: str
 
 
 def battery(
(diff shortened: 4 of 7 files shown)


custom_components/ef_ble/eflib/devices/shp2.py
```diff
@@ -29,11 +30,14 @@ class ControlStatus(IntFieldValue):
     STANDBY = 4
 
 
-class ForceChargeStatus(IntFieldValue):
```
ecammit, line 32: `ForceChargeStatus` was removed as it's a simple binary and the sensor was moved to a binary sensor.

custom_components/ef_ble/number.py
```diff
@@ -295,6 +302,34 @@ class EcoflowNumberEntityDescription[Device: DeviceBase](NumberEntityDescription
...
+        native_unit_of_measurement=PERCENTAGE,
+        native_step=1.0,
+        native_min_value=10,
+        native_max_value=50,
+        async_set_native_value=(
+            lambda device, value: device.set_backup_battery_level(int(value))
+        ),
+        availability_prop="backup_reserve",
+    ),
+    EcoflowNumberEntityDescription[shp2.Device](
```
ecammit, line 317: Unfortunately I had to add a new control for `backup_charge_limit` because the `battery_charge_limit_max` control has `min_value_prop="battery_charge_limit_min"`. So if I used `battery_charge_limit_max`, I'd have to create a `battery_charge_limit_min` property to limit the control to a lower bound of 80. However, there is also a control for `battery_charge_limit_min`, so if I added `battery_charge_limit_min=80`, it'll add another control for `battery_charge_limit_min`.

I think longer-term we probably shouldn't link those controls like that.

GnoX: Yeah, that's the main con of this design. I'm currently working on something that would allow devices to define controls independently straight from the device class. It is going to look something like this:

```python
    @controls.number(
        energy_backup_battery_level,
        min=battery_charge_limit_min,
        max=battery_charge_limit_max,
        availability=energy_backup,
        translation_key="energy_backup_level",
    )
    async def set_backup_reserve_level(self, value: int):
        ppas = pd303_pb2.ProtoPushAndSet()
        value = min(max(10, value), 50)
        ppas.backup_reserve_soc = value

        await self._send_config_packet(ppas)
        return True
```

custom_components/ef_ble/sensor.py
```diff
@@ -359,6 +360,7 @@ def shp2_circuit(
     "remaining_time_discharging": duration(enabled=False),
     # SHP2
     "grid_power": power(precision=1),
+    "power_status": enum(options=shp2.PowerStatus, icon="mdi:home-lightning-bolt"),
```
GnoX: You should add that icon to `icons.json` instead of here.

custom_components/ef_ble/number.py
```diff
@@ -87,13 +89,18 @@ class EcoflowNumberEntityDescription[Device: DeviceBase](NumberEntityDescription
...
+        | delta_pro_3.Device
+        | river2.Device
+        | shp2.Device
     ](
         key="ac_charging_speed",
         name="AC Charging Speed",
         device_class=NumberDeviceClass.POWER,
         native_unit_of_measurement=UnitOfPower.WATT,
         native_step=1,
+        step_value_prop="ac_charging_speed_step",
```
GnoX, line 103: Is this just to reflect app behavior? Does the command fail if the step is not multiple of 100? 

My devices do accept any value within min max range, but app just artificially limits it to these steps.

ecammit: Yes this is to mimic the app. My goal has been to replicate the app constraints as much as possible because I don't trust the Ecoflow firmware to protect me from an expensive misconfiguration.

While I /can/ set any value for SHP2, I'm honestly afraid to leave it at such a setting. The firmware coding by Ecoflow is historically so bad partly because they frequently bake business logic enforcement and input validation into the app UI and MQTT server layers rather than in the unit's operating firmware. I presume because development and testing is far easier for app and server code than it is for embedded circuit firmware.... or they could just have better devs for one over the other.

Regardless, for some examples on just the SHP2, look at the bug we found of how the panel's circuit relays start glitching when you set circuits in a split pair to different values. Another is how both force charge and EPS mode are made explicitly mutually exclusive with operating mode in the app by showing a confirmation modal when changing one value and it confirming that you desire to turn the others off. The MQTT enforces this somewhat, but even that was not reliable in all cases back when I was writing the internet-enabled integration code. However the BLE config set commands happily allow force charge or EPS to be on and operating mode to still be set. Worst case from this might be the grid feed and the battery feed activating on the panel at the same time, but in mis-aligned phases... with lots of arcing and burning.

So while the SHP2 works at any number, they could have enforced an interval of 100w because of something we don't know, like:
1. they found some electrical component regulating the wattage overheats and burns out when set to certain values that happen to not align on an interval of 100
2. there could be rounding or poor data type casting bugs in the firmware's math that goes away when using intervals of 100. The panel limits the charging watts for batteries that are nearly empty or nearly full and limits charging watts below this setting to only 80% of the configured breaker amperage setting of the panel after subtracting out any active loads on the circuits (I presume this is to satisfy the NEC requirements of derating for high-amp continuous loads, like with EV chargers). Any of this math could suddenly change from a result with no decimal to something with a decimal, that could overflow stacks or cause other chaos.

So I would like to keep this as intervals of 100 for safety. But we could certainly add some option in the configuration flow to disable app constraints like this, for if people really want to "live on the edge". Optionally I could just keep a rounding to 100 in the set function. It would allow a user to set it to anything within the range, but it would be rounded to the nearest 100 before sending it to the unit and the ACK from the panel would cause the number slider control to "snap" to the nearest 100. The only concern with this might be if someone has an automation triggered for this number entity. It would trigger twice in quick succession. And if the automation is also trying to set the value, it could result in an infinite loop of re-triggering.

Anyhow, I hope your weekend has been well. Thanks for listening and reviewing!

custom_components/ef_ble/sensor.py
ecammit: I didn't know about icons.json until now, so thanks. I'll migrate the icon selection into that.

custom_components/ef_ble/number.py
GnoX: Thanks for the explanation, I was just curious, did not mean it as a change request.

Though based what I've seen in the app, I don't think that they intentionally limit it based on devices because they don't do it in the business logic layer (there are few exceptions), but they do it from the presentation layer that is not tied to the device and I don't really see a pattern of when they use 1 or 100 step components. But yeah, with high voltage devices, it's better be safe than sorry :)

custom_components/ef_ble/binary_sensor.py
```diff
@@ -114,8 +121,9 @@ def shp2_channel(
     "ch{n}_backup_is_ready": shp2_channel(
         battery, "channel_backup_is_ready", enabled=False
     ),
+    "ch{n}_force_charge": shp2_channel(power, "backup_force_charge", enabled=False),
     # SHP2 energy binary sensors
-    "channel{n}_is_enabled": shp2_channel(power, "channel_is_enabled"),
+    "channel{n}_is_enabled": shp2_channel(power, "channel_is_enabled", enabled=False),
```
ecammit, line 125: Default to disabled, since this is now a duplicate sensor of the switches that were added.

ecammit:
@GnoX, sorry for the re-review, but I pushed a couple minor cleanup changes... moved a few icons I had missed to icons.json and removed some redundant translation_key properties.

GnoX:
No worries, I'm going to be publishing a MR shortly after this one is merged so I would've probably done it myself before release anyways. I'm almost done reworking the whole control registry implementation and simplified definition of those indexed SHP fields.

PR #222, opened 2026-03-13
Clean up sensor registration code

GnoX:
This PR cleans up sensor registration code to be more concise without impacting readability, but making it easier to look through all sensors. It also improves registration of indexed sensors - the way we were doing it previously may potentially allow key overrides that we wouldn't detect - by moving it to static strings and formatting, ruff check does not pass because of duplicate keys.

Additionally, this PR sets initial values from device rather than None default as some devices receive data incrementally and this could cause sensors to be unavailable for a long time until it receives callback.

This is part 1 of a series of pull requests aimed at improving sensors and controls structure and dev experience.

### DIFF
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -1,8 +1,8 @@
 """EcoFlow BLE binary sensor"""
 
-import logging
 from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


GnoX:
@rabits Noticed an error with enums here, if you were planning a release, don't do it yet.

PR #225, opened 2026-03-15
Fix Wave 3 power switch not working

pattex007:
Two issues prevented the Wave 3 power on/off switch from functioning:

1. Config packets were sent to wrong destination address (0x14 instead of 0x42). The Wave 3 receives data from src=0x42, so config packets must target dst=0x42, matching the pattern used by all other protobuf-based devices.

2. enable_power() used cfg_sys_pause=False to turn on, but in proto3 the default value for bool (false) is not serialized, resulting in an empty message that the device ignores. Replaced with cfg_power_on=True / cfg_power_off=True which always serialize correctly.

### DIFF
--- a/custom_components/ef_ble/eflib/devices/wave3.py
+++ b/custom_components/ef_ble/eflib/devices/wave3.py
@@ -152,11 +152,9 @@ class Device(DeviceBase, ProtobufProps):
         return True
 
     async def enable_power(self, enabled: bool):
+        cfg = ac517_apl_comm_pb2.ConfigWrite()
         if enabled:
-            await self._send_config_packet(
-                ac517_apl_comm_pb2.ConfigWrite(cfg_power_on=True)
-            )
+            cfg.cfg_power_on = True
         else:
-            await self._send_config_packet(
-                ac517_apl_comm_pb2.ConfigWrite(cfg_power_off=True)
-            )
+            cfg.cfg_power_off = True
+        await self._send_config_packet(cfg)


custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -124,7 +124,7 @@ async def data_parse(self, packet: Packet):
 
     async def _send_config_packet(self, message: ac517_apl_comm_pb2.ConfigWrite):
         payload = message.SerializeToString()
-        packet = Packet(0x20, 0x14, 0xFE, 0x11, payload, 0x01, 0x01, 0x13)
+        packet = Packet(0x20, 0x42, 0xFE, 0x11, payload, 0x01, 0x01, 0x13)
```
GnoX, line 127: So, did the battery limit commands not work either?

I remember finding those headers somewhere, but I may have got them confused for something else.

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -152,7 +152,11 @@ async def set_battery_charge_limit_max(self, limit: int):
...
-            ac517_apl_comm_pb2.ConfigWrite(cfg_sys_pause=not enabled)
-        )
+        if enabled:
+            await self._send_config_packet(
+                ac517_apl_comm_pb2.ConfigWrite(cfg_power_on=True)
+            )
+        else:
+            await self._send_config_packet(
+                ac517_apl_comm_pb2.ConfigWrite(cfg_power_off=True)
+            )
```
GnoX:
```suggestion
        cfg = ac517_apl_comm_pb2.ConfigWrite()
        if enabled:
            cfg.cfg_power_on = True
        else:
            cfg.cfg_power_off = True
        await self._send_config_packet(cfg)
```

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -152,7 +152,11 @@ async def set_battery_charge_limit_max(self, limit: int):
         return True
 
     async def enable_power(self, enabled: bool):
```
GnoX, line 154: So, I wasn't the one who implemented this command, I think it was a different contributor that had  Wave 3 and used AI so I believed it was working correctly. Did that `cfg_sys_pause` do nothing even after you fixed headers?

GnoX (changes requested):

custom_components/ef_ble/eflib/devices/wave3.py
pattex007: No, with firmware 1.0.0.92, `cfg_sys_pause` does nothing. I've checked this twice. Several other features are also not working, such as fan level, water level (fixed at 0), and indoor return air temperature. There may have been breaking changes with this firmware version.

GnoX: Okay, I didn't bother really checking the updated implementation and believed it worked as it was implemented by someone owning a Wave 3, but looking at the implementation, some sensors never could've worked - it uses message that is only sent over mqtt, not bluetooth.

Feel free to list all sensors that don't work, we'll remove them. Or you can remove them yourself as a part of this PR.

PR #237, opened 2026-03-21
Add climate entity support and enhance Wave 3 device features

pattex007:
- Introduced EcoFlow BLE climate entity for Wave 3, enabling HVAC mode and temperature control.
- Added fan speed and temperature display type options for Wave 3 devices.
- Updated README to reflect new features and entity types.
- Enhanced Wave 3 device properties with additional power metrics and state management.
- Added translations for new features in the English language file.


custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -115,11 +157,15 @@ async def data_parse(self, packet: Packet):
             )
             processed = True
 
+        if self.sleep_state is not None:
+            self.power = self.sleep_state == SleepState.ON
```
GnoX, line 161: Why did you change this? This should do the same thing as that pb field.

custom_components/ef_ble/climate.py
```diff
@@ -0,0 +1,180 @@
+"""EcoFlow BLE climate entity for Wave 3."""
```
GnoX: This will not be only be for wave 3 but for other climate devices, this docstring is wrong.

custom_components/ef_ble/climate.py
```diff
@@ -0,0 +1,180 @@
...
+    hass: HomeAssistant,
+    entry: DeviceConfigEntry,
+    async_add_entities: AddEntitiesCallback,
+) -> None:
+    device = entry.runtime_data
+    if isinstance(device, wave3.Device):
+        async_add_entities([EcoflowClimateEntity(device)])
+
+
+class EcoflowClimateEntity(EcoflowEntity, ClimateEntity):
```
GnoX, line 93: This uses completely different style of entity registration from our other entities. It should use entity descriptions for this and some of the attributes should be based on properties, i.e. min and max temps, available fan and hvac modes.

custom_components/ef_ble/climate.py
```diff
@@ -0,0 +1,180 @@
...
+        self.async_write_ha_state()
+
+    @callback
+    def _on_current_temperature(self, state):
+        self.async_write_ha_state()
+
+    @callback
+    def _on_fan_speed(self, state):
+        self.async_write_ha_state()
+
```
GnoX: What's the point of these callbacks? They do the exact same things.

custom_components/ef_ble/eflib/devices/wave3.py
pattex007: You are right, the result is the same.

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -151,10 +197,117 @@ async def set_battery_charge_limit_max(self, limit: int):
...
+        )
+
+    async def set_target_temperature(self, temperature: float):
+        await self._send_config_packet(
+            ac517_apl_comm_pb2.ConfigWrite(cfg_temp_set=temperature)
+        )
+        self.target_temperature = round(temperature, 1)
+        self._update_cached_mode_temp(temperature)
+        self.update_callback("target_temperature")
+        self.update_state("target_temperature", self.target_temperature)
```
GnoX, line 276: I don't think we should be calling callbacks from here, the device should update it from the fields as we don't know if the command worked. This could potentially break the state.

custom_components/ef_ble/__init__.py
```diff
@@ -144,6 +145,27 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
...
+                dev_reg.async_update_device(
+                    dev_entry.id, sw_version=str(version)
+                )
+
+        if device.sw_version is not None:
+            _update_sw_version(device.sw_version)
+        else:
+            device.register_state_update_callback(
+                _update_sw_version, "pd_firm_ver"
+            )
```
GnoX: I don't know if we want to do this here. Can you remove it and we'll readd it in a different MR? It's going to be hard to modify if we do it right now.

README.md
```diff
@@ -356,27 +356,30 @@ first charging task</sup>
...
+| Battery Power                    |               |                 |                        |
+| Indoor Supply Air Temperature    |               |                 |                        |
+| Indoor Return Air Temperature    |               |                 |                        |
+| Outdoor Ambient Temperature      |               |                 |                        |
+| Condenser Temperature            |               |                 |                        |
+| Evaporator Temperature           |               |                 |                        |
+| Compressor Discharge Temperature |               |                 |                        |
+
+> **📝 Note:** Wave 3 also provides a **Climate** entity with HVAC modes
+> (Cool, Heat, Fan, Dry, Auto), target temperature, and fan speed control.
```
GnoX: Should not be a note but an another column.

custom_components/ef_ble/sensor.py
```diff
@@ -591,21 +591,30 @@ def shp2_circuit(
...
+    "ac_input_power": power(precision=1),
+    "battery_power": power(precision=1),
+    "module_wifi_rssi": EcoflowSensorEntityDescription(
+        key="",
+        native_unit_of_measurement="dBm",
+        device_class=SensorDeviceClass.SIGNAL_STRENGTH,
+        state_class=SensorStateClass.MEASUREMENT,
+        entity_category=EntityCategory.DIAGNOSTIC,
+        entity_registry_enabled_default=False,
+    ),
```
GnoX, line 603: Add another description factory `signal` that does this, do not use descriptions directly here.

GnoX:
Well, I mean this is clearly an AI assisted PR so I'm not sure how well you're be able to respond to all of my comments. There are still some things that are not ideal that I didn't comment yet about because I have to look deeper on how the data looks. I don't want this to take up too much time, AI may not resolve those issues correctly for several attempts so one suggestion is to let me take over this and let you test my changes?

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -151,10 +197,117 @@ async def set_battery_charge_limit_max(self, limit: int):
...
+    def _extract_mode_params(self, msg: ac517_apl_comm_pb2.DisplayPropertyUpload):
+        if msg.HasField("wave_mode_info"):
+            self._cached_mode_items = [
+                {
+                    "submode": item.submode if item.HasField("submode") else None,
+                    "airflow_speed": item.airflow_speed if item.HasField("airflow_speed") else None,
+                    "temp_set": item.temp_set if item.HasField("temp_set") else None,
+                }
+                for item in msg.wave_mode_info.list_info
+            ]
```
GnoX, line 209: What's the purpose of this caching? This looks quite unnecessary.

pattex007:
> Well, I mean this is clearly an AI assisted PR so I'm not sure how well you're be able to respond to all of my comments. There are still some things that are not ideal that I didn't comment yet about because I have to look deeper on how the data looks. I don't want this to take up too much time, AI may not resolve those issues correctly for several attempts so one suggestion is to let me take over this and let you test my changes?

Yes, we can certainly do it that way. I've implemented some of your change requests.

custom_components/ef_ble/select.py
```diff
@@ -155,6 +156,27 @@ class EcoflowSelectEntityDescription[T: DeviceBase](SelectEntityDescription):
...
+    EcoflowSelectEntityDescription[wave3.Device](
+        key="fan_speed",
+        name="Fan Speed",
+        options=wave3.FanSpeed.options(include_unknown=False),
+        set_state=(
+            lambda device, value: device.set_fan_speed(
+                wave3.FanSpeed[value.upper()]
+            )
+        ),
+    ),
```
GnoX, line 179: This will probably break Wave 2 switch - there's a fan_speed entity registered here.

pattex007: That's still an open question; actually, a separate fan speed control isn't necessary, since that works via the climate control, right?

custom_components/ef_ble/eflib/devices/wave3.py
pattex007: Without caching, depending on the device's responses, I didn't have correct data.

GnoX: You should be able to define all of these as fields though, we're losing a lot of typing checks this way.

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -99,12 +132,21 @@ def check(cls, sn):
...
         return Packet.fromBytes(data, xor_payload=True)
 
+    async def request_full_upload(self):
+        await self._send_config_packet(
+            ac517_apl_comm_pb2.ConfigWrite(
+                active_display_property_full_upload=True,
+                active_runtime_property_full_upload=True,
+            )
+        )
+
```
GnoX, line 142: Can you send me diagnostics info with enabled diagnostics collection and this full upload enabled? I wasn't able to make this work for bluetooth so I'm a bit suspicious whether this really works.

GnoX: Thanks for the diagnostics, but I think that it does not include the response of this if it does something. I think what this is supposed to do is send a single response packet with configuration, but I'm not sure myself, I couldn't figure out how this works. There is nothing in your data to suggest that this had any effect.

pattex007: Without this function, I have no current data on the status of, for example, the desired temperature. I will send the diagnostics from a longer period. Perhaps that will help.

GnoX: Can you send data from when you didn't send this request? I have some wave3 data from when I was initially implementing this and see no difference between your data and the one from original implementation. It is possible that firmware started omitting it, but it is unlikely.

GnoX:
Okay, cleaning up these AI PR's is making me work a little bit more than writing it from scratch. I cleaned everything up in #238 and changed few things around. 

@pattex007 Can you try if everything works in [v0.38.0-d.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.1)? Can you also download some diagnostics from that version? Make sure that it is running at least 1 minute before downloading it.

GnoX:
@pattex007 Did you try installing that version? I'm going to need that diagnostics to proceed.

pattex007:
Sorry for my late reply, unfortunately it's not working. I'm not getting a correct status update from the device. When I change settings on the device itself, they aren't updated in real time.

GnoX:
Closing - superseded by #238

PR #238, opened 2026-03-22
Add climate entity support and enhance Wave 3 device features

### LINKED ISSUE #285: [Bug]: Wave_3 Support
### What happened?

Please add Wave 3 support to control and view all data , it has been a year of waiting 

latest

latest

Internal bluetooth adapter

_No response_

```text

```

### Relevant log output

```shell

```

GnoX:
This PR implements implements climate entity for Wave 3 and adds some additional sensors.

Original implementation by @pattex007, supersedes #237.


driedler:
Thanks for working on this feature! It's great to finally see HS support for the Wave3. Any idea when this will be merged?

GnoX:
As soon as I get a confirmation that it works and get diagnostics data for testing. However, the original author is not responding to my messages https://github.com/rabits/ha-ef-ble/pull/237#issuecomment-4106469028. 

@driedler If you install version [v0.38.0-d.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.1) and upload some [diagnostics](https://github.com/rabits/ha-ef-ble/wiki/Providing-Diagnostics-Data-for-Debugging-or-Implementing-New-Sensors) here, we can continue.

GnoX:
@driedler @pattex007 Can you try [v0.38.0-d.2](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.2)? I think climate should be working there.

pattex007:
> @driedler @pattex007 Can you try [v0.38.0-d.2](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.2)? I think climate should be working there.

Unfortunately, it doesn't work.

BenjiGorilla:
> As soon as I get a confirmation that it works and get diagnostics data for testing. However, the original author is not responding to my messages [#237 (comment)](https://github.com/rabits/ha-ef-ble/pull/237#issuecomment-4106469028).
> 
> @driedler If you install version [v0.38.0-d.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.1) and upload some [diagnostics](https://github.com/rabits/ha-ef-ble/wiki/Providing-Diagnostics-Data-for-Debugging-or-Implementing-New-Sensors) here, we can continue.

I've tried this version. Seems to be mostly working. Some weird behaviour in certain mode though (setting temperature in auto mode doesn't seem to work). Setting temperature in heat and cool modes does work correctly. 

Fan speed settings are listed as "Low, medium_low, Medium, medium_high, High". List items staring with "medium_" don't have icons assigned. - Minor quality issues with the labeling, but otherwise functions correctly.

It's not possible to select max, eco, and sleep modes.

[config_entry-ef_ble-01KPG6F4GMZ1YAQGAN0YACQ1HJ.txt](https://github.com/user-attachments/files/26876221/config_entry-ef_ble-01KPG6F4GMZ1YAQGAN0YACQ1HJ.txt)

BenjiGorilla:
Thanks for working on supporting this device :)

GnoX:
Ah, I was looking through the code and couldn't identify what could be wrong with that climate control - @pattex007 are you sure you were running on the correct version? Or by `it doesn't work` you meant that what as not working were the things @BenjiGorilla mentioned?

@BenjiGorilla Thanks - I'll attempt to fix those issues soon.

GnoX:
@BenjiGorilla Just released [v0.38.0-d.3](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.3) that should fix that auto mode temps (but it was renamed to heat/cool to better work in HA) and also added select for sub mode, though I'm not sure how those sub modes work - are they available for all of the main modes or only for some of them?

BenjiGorilla:
> @BenjiGorilla Just released [v0.38.0-d.3](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.3) that should fix that auto mode temps (but it was renamed to heat/cool to better work in HA) and also added select for sub mode, though I'm not sure how those sub modes work - are they available for all of the main modes or only for some of them?

Tested the new release. Really good work thanks :1st_place_medal: ! Gives me everything I need with the exception of the first point below (as I like to use the dehumidifier mode). Everything else I mention is just info from my testing to make the implementation perfect, but could be looked at in a later release as I appreciate dev time is a finite resource :)

1. The only major issue is that in Dry mode it is not possible to set the humidity target level. HA still displays a temperature selector, which is irrelevant in this mode. When switching to dry mode the wave 3 defaults to the last selected humidity target.

2. Minor issue: in fan only mode HA displays a temperature slider, but this is irrelevant in this mode. 

The device is still new to me, so I'm still learning all it's features.

By ecoflows design:
- Submodes are only available in heat and cool mode.
- Fans speeded selection is available in all main modes except auto mode (heat/cool mode in HA).

When selecting an invalid combination HA will change the value back after a few seconds. So not the end of the world.

Add on battery:
My wave 2 addon battery does not support charging limits. My wave 3 add on battery does support this function. I'm not sure if you can detect which type of add on battery is connected via BLE? The ecoflow app can figure this out somehow, as it hides the charge limit options when the wave 2 addon battery is connected. I had the wave 2 add on battery connected during my diagnostic run, so maybe I need to do this again with the wave 3 add on battery fitted?

GnoX:
@BenjiGorilla Try [v0.38.0-d.4](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.4) - it should adress everything except the add on battery.
For add-on battery, I'm going to need those diagnostics with Wave 3 battery to see what field to use for identification.

BenjiGorilla:
> @BenjiGorilla Try [v0.38.0-d.4](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.4) - it should adress everything except the add on battery. For add-on battery, I'm going to need those diagnostics with Wave 3 battery to see what field to use for identification.

I've tried the new version and all seems to be working well thank you! I'm away currently, but will get diagnostics for the add-on battery this weekend.

almolook:
Screenshot 2026-05-01 at 8 14 01 AM
v0.38.0-d.4  works fine now, expect the temperature, it shows the target temperature as 15.8 while I set it to 16
and when I increase it to 18 , it show 17.8

GnoX:
@almolook Thanks for feedback, [v0.38.0-d.5](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.38.0-d.5) should make temperatures snap to the closest step rather than adding it to the current target value when using `+` and `-` buttons. BTW, you can remove all of those greyed out sensors manually, they won't be coming back as they are not available over bluetooth.

custom_components/ef_ble/eflib/devices/wave3.py
```diff
@@ -72,25 +100,48 @@ class Device(DeviceBase, ProtobufProps):
...
+    ac_input_power = pb_field(pb_disp.pow_get_ac, pround(1))
+    battery_power = pb_field(pb_disp.pow_get_bms, pround(1))
 
     temp_indoor_supply_air = pb_field(pb_disp.temp_indoor_supply_air, pround(1))
-    temp_indoor_return_air = pb_field(pb_run.temp_indoor_return_air, pround(1))
-    temp_outdoor_ambient = pb_field(pb_run.temp_outdoor_ambient, pround(1))
-    temp_condenser = pb_field(pb_run.temp_condenser, pround(1))
-    temp_evaporator = pb_field(pb_run.temp_evaporator, pround(1))
-    temp_compressor_discharge = pb_field(pb_run.temp_compressor_discharge, pround(1))
+    # temp_indoor_return_air = pb_field(pb_run.temp_indoor_return_air, pround(1))
```
rabits: Don't really like to keep commented code in the sources. Let's leave it for the git history.

custom_components/ef_ble/eflib/devicebase.py
```diff
@@ -422,14 +425,28 @@ def remove_state_update_callback(
         """Remove previously registered state update callback"""
         self._state_update_callbacks[propname].discard(callback)
 
-    def update_state(self, propname: str, value: Any):
+    def update_state(self, propname: "str | Field[Any]", value: Any):
```
rabits: Sorry maybe I'm a bit outdated, but why is multi-type in quotes? Isn't it against the rules of syntax validation to keep such control structures in a string? See that multiple times across the logic...

rabits (changes requested):
Hmmm, small things...

custom_components/ef_ble/eflib/devicebase.py
GnoX: Python does interpret string literals in type hints as regular types and you generally use it for forward declarations or when using `if TYPE_CHECKING` import guards. I'm not sure why I used in this case though, it should work without it, I'll change it.

custom_components/ef_ble/eflib/devices/wave3.py
GnoX: I do tend to agree, but in some cases, I want to keep it for later to show that it may be possible implement these sensors, otherwise we'd just forget it or have to retrace it again. But looking at these sensors, this does not apply here, none of these are possible to do anyway, I'll remove it.

custom_components/ef_ble/eflib/devicebase.py
rabits: Ok, sorry did not marked those as well: noticed https://github.com/rabits/ha-ef-ble/pull/238/changes#diff-8c0fea78b94b234ce496d8e18082fb89f67daee5beac2f5425178f4360d88dfcR322 - is it ok to have return values described in string and not as type? I see ` -> "Type"` pattern in git grep, but also just regular ` -> Type...` what's the deal there?

custom_components/ef_ble/eflib/devices/wave3.py
rabits: Thank you! Not sure how to address the unimplemented fields - but I think it's better just to keep the clean logic. Anyway anyone who would want to implement a new field need to know it's there and check the pb name for it (because it could change in time). Comments are not good to keep long-term, because we have no rules to verify them (too complicated, so they've become misleading), so better to omit entirely.

custom_components/ef_ble/eflib/devicebase.py
GnoX: String types are equivalent to plain types in typing context so it does not impact dev experience. Also, python 3.14 implemented [deferred annotations](https://docs.python.org/3/whatsnew/3.14.html#pep-649-pep-749-deferred-evaluation-of-annotations) so string annotations are no longer needed, but there may be some people that are still using lower pythons so we don't want to remove all occurences just yet. 

I did generate the code using AI so I might miss some non-crucial things, but I'm reviewing and refining it to fit the style and design. I find it to be a great tool for experiences devs and quick apps, but a nightmare to review if coming from less experienced devs. Anyways, I removed all quoted types from places where they're not required in 31a77d9

rabits: Thank you!)

rabits (approved):
Looks good

PR #267, opened 2026-04-03
Relax protobuf requirements

GnoX:
This PR lowers the limit for protobuf requirement to `~=6.30`, we didn't notice that [HA package requirements](https://github.com/home-assistant/core/blob/86d80c96d732036f085e15679c25acfebbaa15a3/homeassistant/package_constraints.txt#L152) do have a strict pin of  `==6.32.0` making conflicts.


rabits (approved):
Yep, agree

Issue #28, opened 2025-05-08
expose DPU and SHP2 entities


MatthewJB2022:
Hello and thanks again for your work on this.   would it be possible to expose the following entities on the DPU

- power in/out voltage, amp, watts, hrz
-solar volt, amp, watts for both HI and LOW Pv
-individual battery percentages

and on the SHP2
-charging limits

GnoX:
Hi, I was updating code for DPU and SHP2 today to the new style and tried to add some of the entities you mentioned - can you try installing version [v0.12.0-a.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.12.0-a.1) from my fork? I don't have any of these devices so it is mostly untested and there is high possibility that I made an error somewhere. If it works though, it'll be relatively easy adding all of the remaining sensors.

MatthewJB2022:
Hi - sure i'll try this now

MatthewJB2022:
hi - i tried this but it still shows just 5 entities for the DPU.  no new changes

GnoX:
Are you certain you installed the pre-release version I mentioned and disabled integration from this repo? It should at least show up those sensors as unknown (or crash).

MatthewJB2022:
possibly not, as i'm really not good at this stuff :).   here's what i did.    i deleted all of the DPU/shp2 devices i had in HA and then the integration went away.   then i did a new custom component installation in HACS using the github link you provided me with the new version.   installed it and went about installing the DPU device and got the same old results.   what did i do wrong?

GnoX:
I think you're almost there, but installed the latest version instead of a pre-release version. Just pasting in the link won't work, you have to click on the HACS integration, click on three dots in top right corner, select redownload and select the latest pre-release version and restart HA.

![Image](https://github.com/user-attachments/assets/29572ea8-2355-4ede-9c57-a211efb6ccae)

MatthewJB2022:
oh crap ok let me try this hang on

MatthewJB2022:
ok - seems to be on right track.   i now have 14 entities (not 5) for a dpu...although several of them say unknown like the 2 solar current, 2 solar voltage and obviously the extra battery slots (since i only have 2).   as for the shp2 it went from 31 to now 30 entities.

MatthewJB2022:
![Image](https://github.com/user-attachments/assets/9c923405-4360-4753-94fa-59eb6c0d8b7b)

MatthewJB2022:
[...]

PR #269, opened 2026-04-03
DPU Bluetooh packet chatter reduction:

ecammit:
Testing shows the DPU does not need the reply packets to function properly:
- Removes unneccessary reply packets

This should help reduce packets piling up in queue for slower connections:
- Moves BackendRecordHeartbeatReport request from a timed task to queue on receipt of a AppShowHeartbeatReport message

### DIFF
--- a/custom_components/ef_ble/eflib/devices/dpu.py
+++ b/custom_components/ef_ble/eflib/devices/dpu.py
@@ -1,5 +1,4 @@
 from enum import IntEnum
-from functools import partial
 
 from ..commands import TimeCommands
 from ..devicebase import AdvertisementData, BLEDevice, DeviceBase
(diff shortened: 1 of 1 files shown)


GnoX (approved):

rabits:
Hmmm, this one is a bit complicated - in theory DPU can give much more information (like described in #28) - but unfortunately I was unable to find the right mixture for DPU (for SHP2 replies actually help to get the additional data, so maybe for DPU there is something else). But I think since additional data is not working - replies are not needed right now.

rabits (approved):
Agree, but maybe in the future it will be needed for #28.

ecammit:
@rabits That additional data from #28 is already added, it's just disabled in the Diagnostics section by default. The Labs screen data is delivered in the `BackendRecordingHeatbeatReport` messages that is triggered by the `self.request_heartbeat_info(8)` call. I had to go to the decompiled app source to figure out how the Labs screen triggers it. Unfortunately even in the app it is a message that has to be requested every time. The DPU doesn't automatically send it. And honestly that may be a deliberate choice from EcoFlow to prevent that extra data from sending to the MQTT servers when it's rarely ever needed.

I was previously requesting the `BackendRecordingHeatbeatReport` using a timed task, but in this PR I moved it to make a request every time an `AppShowHeartbeatReport` is received. This way if the connection is slow, the timed task doesn't keep firing and stacking up requests. The frequency at which `AppShowHeartbeatReport` is received is a good enough frequency to get the `BackendRecordingHeatbeatReport` values.

PR #282, opened 2026-04-09
Migrate Delta Pro 3 to controls system and add several diagnostics sensors

### LINKED ISSUE #274: [Feature request] - add Error code & BMS Run state
Out of the blue, The AC out of my Delta Pro 3 (EU version - no LV AC out, just HV) turned off, in spite of being set to `Never` in the timing settings.
I am running in parallel for now **your** integration and another one side by side.
Looking at the history, I found that there was an `302` error code thrown by the DP3 and that the `BMS Run state` turned to `0` before the AC went off. Both were reported by the other integration.
Would be nice to have these parameters available **in your** integration. Would help me getting rid of the other integration - yours is such a relief ...

LATER EDIT: Another very useful add-on would be the **Operating Mode** (a.k.a **Energy Strategy Mode**) control with its 3 values; 
- **Off** or, better, **Default**, if you like - less confusing, 
- **TOU** and 
- **Self-Powered** 

Possibly as a drop-down list?
Currently I have to set this from the official app. 
Had to do this when the error was thrown - normally I use `Self-Powered` but it was reverted to `Default`, probably by the error handling routine within the DP3 firmware, most likely.

GnoX:
This PR migrates Delta Pro 3 to controls system support ands some diagnostic sensors - BMS Run State and Error code that are disabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,8 @@ Click on any device below to see available sensors, switches, and controls:
 | Cell Temperature (disabled)         |                |                      |
 | Charge Time Remaining (disabled)    |                |                      |
 | Discharge Time Remaining (disabled) |                |                      |
+| Error Code (disabled)               |                |                      |
+| BMS Run State (disabled)            |                |                      |
 
 <sup>¹ Per extra battery (up to 2)</sup>
 
--- a/README.md
+++ b/README.md
@@ -302,10 +302,11 @@ Click on any device below to see available sensors, switches, and controls:
 | Cell Temperature (disabled)         |                |                      |
 | Charge Time Remaining (disabled)    |                |                      |
 | Discharge Time Remaining (disabled) |                |                      |
-| Error Code (disabled)               |                |                      |
+| Error Occurred (disabled) ²         |                |                      |
 | BMS Run State (disabled)            |                |                      |
 
 <sup>¹ Per extra battery (up to 2)</sup>
+<sup>² Includes error code as an extra attribute</sup>
 
 > **📝 Note:** Extra batteries are detected when the integration is set up. If you
 > connect or disconnect a battery, you can reconfigure the integration to update
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -122,6 +122,7 @@ def shp2_channel(
 _BINARY_SENSORS: Final[dict[str, BinarySensorEntityDescription]] = {
     "error_happened": problem("error", entity_category=EntityCategory.DIAGNOSTIC),
     "plugged_in_ac": plug(),
+    "bms_run_state": power(enabled=False, entity_category=EntityCategory.DIAGNOSTIC),
     # SHP2 backup channel binary sensors
     "ch{n}_backup_is_ready": shp2_channel(
         battery, "channel_backup_is_ready", enabled=False
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -1,8 +1,8 @@
 """EcoFlow BLE binary sensor"""
 
 from collections.abc import Callable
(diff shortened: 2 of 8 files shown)


GnoX:
## Code Review

### Overview
Two main changes:
1. **Controls migration** — adds `@controls.switch`, `@controls.outlet`, `@controls.battery`, `@controls.power` decorators to all Delta Pro 3 command methods, converting from manual entity registration to the declarative controls system
2. **Diagnostics sensors** — adds `error_code`, `bms_run_state`, and `error_occurred` to Delta Pro 3, Delta 3 family, and River 3. Introduces `state_attribute_fields` support in `EcoflowBinarySensorEntityDescription` so `error_occurred` can expose `error_code` as an extra attribute

### Looks good
- Controls migration follows the established pattern from other devices (River 3, Delta 3, etc.)
- `int` → `float` param types with `int()` casts are consistent with how HA number entities deliver values
- `solar_lv_power`/`solar_hv_power` converted from `Field` + manual assignment to `@computed_field` — cleaner, and `_notify_updated()` replaces the manual callback loop
- `error_occurred` as a `@computed_field` derived from `error_code` is the right pattern
- `state_attribute_fields` on `EcoflowBinarySensorEntityDescription` is a clean, reusable mechanism for any binary sensor that needs extra attributes
- Icons and translations complete for all new entities

### Issues

- **`error_occurred` defined in both `_delta3_base.py` AND `delta_pro_3.py`** — Delta Pro 3 inherits from `DeviceBase` (not `Delta3Base`), so the `_delta3_base` definition gives it to all Delta 3 variants, and `delta_pro_3` adds its own. That's correct and not a conflict, but the DP3 one shadows the Delta 3 base one if DP3 ever inherits from Delta3Base. Worth a comment noting DP3 has its own because it doesn't inherit from Delta3Base.

- **`_flow_is_on` still duplicated in `delta_pro_3.py`** — this function is identical to `flow_is_on` in `_delta3_base.py` (and the new `transforms.py` in PR #289). If #289 lands first, this will need a rebase. Not blocking, just a merge order note.

### Suggestions

- **`extra_state_attributes` return type** — returning `dict | None` works, but `{} or None` evaluates to `None` when all attribute fields happen to be missing via the `hasattr` check. If `state_attribute_fields` is configured but the device doesn't have those attributes *yet* (before first data parse), the property returns `None` which is fine. But after first parse when `error_code` exists, it'll return `{"error_code": <value>}`. This is correct behavior, just noting the empty-dict → None coercion is intentional.

- **`bms_run_state` is DP3-only but registered globally** — it's defined as a `pb_field` only on `delta_pro_3.Device`, but the `_BINARY_SENSORS` dict entry for `bms_run_state` is global. The `hasattr(device, sensor)` check in `async_setup_entry` handles this correctly, but it means the key exists in the registry for all devices even though only DP3 has the attribute. This is the existing pattern throughout the codebase, so consistent — just noting it.

- **README footnote ²** — says "Includes error code as an extra attribute" which is accurate but doesn't mention `bms_run_state` is a separate binary sensor. The table layout makes this clear, but the footnote could be misread as applying to both rows.

### Verdict
Clean migration. Controls decorators follow the established pattern, diagnostics sensors are properly wired with icons/translations, and the `state_attribute_fields` mechanism is a good reusable addition. Ship it.

rabits (approved):
Looks good!

GnoX:
@rabits I started doing those AI reviews locally some time ago, but figured that it sumarizes and points out issues quite well so I'll be including them in my PRs if you think it's a good info.

rabits:
@GnoX no issues if you find that useful - I prefer to look at the code)

GnoX:
Yeah, I didn't mean for it to replace reviews, just to supplement it. I find it useful for a quick glance before going into a full review. But if it's just a unnecessary noise for you, there's no point in including it here - I use it locally anyways.

PR #283, opened 2026-04-09
Migrate STREAM devices to controls system and improve task controls

### LINKED ISSUE #278: Charging and discharging ecoflow stream ultra x
Would it be possible to have a charge and discharge toggle/switch

At the moment im using a schedule in the EcoFlow app to start charging at 9am and start discharging at 5pm.

I cant see a way I can completely control the charge and discharge through Home Assistant.

GnoX:
This PR migrates all STREAM battery devices to the controls system and adds charging/discharging task controls.

Most noteable changes:
- Added Charging Task and Discharging Task enable/disable switches and Discharging Power Limit slider for STREAM devices
- Added `_send_timer_task_packet` helper with `asyncio.Lock` and pending modification tracking to prevent concurrent commands from overwriting each other
- Added `call_later(delay, callback, key=None)` utility to `Connection`/`DeviceBase` for debounced callbacks with auto-cleanup on disconnect

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,27 +358,27 @@ Click on any device below to see available sensors, switches, and controls:
 
 <br>
 
-| *Sensors*                           | *Switches*           | *Sliders*                    | *Selects*       |
-|-------------------------------------|----------------------|------------------------------|-----------------|
-| Battery Level                       | Feed Grid            | Feed Grid Power Limit        | Energy Strategy |
-| Main Battery Level                  | AC (1) ᴬᴹᴾᵁ          | Backup Reserve Level         |                 |
-| Battery Power                       | AC (2) ᴬᴾᵁ           | Charge Limit                 |                 |
-| Grid Power                          | Charging Task ²      | Discharge Limit              |                 |
-| Grid Voltage                        | Discharging Task ³   | Base Load Power ¹            |                 |
-| Grid Frequency                      |                      | Grid Input Power Limit       |                 |
-| Load from Battery                   |                      | Charging Power Limit ²ꜝ      |                 |
-| Load from Grid                      |                      | Charging Target SOC ²ꜝ       |                 |
-| Load from PV ᴹᴾᵁ                    |                      | Discharging Power Limit ³ꜝ   |                 |
-| AC (1) Power ᴬᴹᴾᵁ                   |                      |                              |                 |
-| AC (2) Power ᴬᴾᵁ                    |                      |                              |                 |
-| PV (1) Power ᴹᴾᵁ                    |                      |                              |                 |
-| PV (2) Power ᴹᴾᵁ                    |                      |                              |                 |
-| PV (3) Power ᴾᵁ                     |                      |                              |                 |
-| PV (4) Power ᵁ                      |                      |                              |                 |
-| PV Power Total ᴹᴾᵁ                  |                      |                              |                 |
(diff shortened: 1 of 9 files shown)


GnoX:
## Code Review

### Overview
Three main areas:
1. **Controls migration** — all STREAM device command methods get `@controls` decorators, with AC outlet overrides in subclasses (`stream_ac_pro`, `stream_max`, `stream_pro`)
2. **Charging/discharging task controls** — new `DischargingTimerTask`, `charging_task_enabled`, `discharging_task_enabled` switches and `discharging_power_limit` slider
3. **Concurrency infrastructure** — `_TimerTaskChain` with `asyncio.Lock` + pending mods for cross-device command safety, and `call_later` utility on `Connection`/`DeviceBase`

### Looks good
- `_send_timer_task_packet` correctly solves the race condition: lock serializes sends, pending mods accumulate and get re-applied on every send, `call_later` clears them after 5s
- `_TimerTaskChain` shared via `ClassVar` dict keyed by `frozenset` of device SNs — clean cross-device coordination without globals
- AC outlet overrides in subclasses are correct: `stream_ac_pro` has both AC 1+2, `stream_max` only AC 1, `stream_pro` only AC 2
- `computed_field` conversion for all derived state fields eliminates the manual `data_parse` boilerplate
- `call_later` in `connection.py` handles key deduplication, connection guard, and auto-cleanup on disconnect
- README updated with new controls and footnotes simplified (removed the "wait 1 second" warning since the lock handles it now)

### Issues

- **`call_later` uses `asyncio.get_event_loop()`** — this is deprecated since Python 3.10 and will warn/fail in newer versions. Should use `asyncio.get_running_loop()` instead. Two occurrences in `connection.py`.

- **`_get_chain()` discovers chain on first call, then caches** — if `_charging_task` is initially `None` (before first data parse), `_get_chain()` falls back to `frozenset([self._sn])` (single-device chain). If `_charging_task` later populates with multiple device SNs, the device is stuck in the wrong chain. This only matters if the first command fires before any data arrives from the device — unlikely in practice, but worth a defensive comment or lazy re-evaluation.

- **`pending_mods` cleanup key is shared** — `key="pending_task_mods"` is the same for both charging and discharging task modifications. If a charging task mod fires, schedules the 5s clear, and then a discharging task mod fires within 5s, the second `call_later` cancels the first timer and resets the 5s window. This is actually correct behavior (extends the window while commands are still in-flight), but the shared key means both types share one cleanup timer. If they should clear independently, use separate keys.

### Suggestions

- **`enable_ac_1`/`enable_ac_2` removed from base class** — the base `stream_ac.Device` previously had these methods, now they only exist in subclasses. If any code calls `device.enable_ac_1()` on a base STREAM AC device (which doesn't have `ac_1` field), it will `AttributeError`. This is likely correct since only Pro/Max/Pro variants have AC outlets, but worth confirming no existing automations target the base device's AC methods.

- **`_send_config_packet` serializes before setting `cfg_utc_time`** — line order is `payload = message.SerializeToString()` then `message.cfg_utc_time = round(time.time())`. The timestamp is set *after* serialization, so it's never included in the payload. This is pre-existing (not introduced by this PR) but worth flagging.

### Verdict
Well-structured PR. The concurrency solution with lock + pending mods + `call_later` cleanup is solid. The `asyncio.get_event_loop()` deprecation should be fixed. Ship it after that.

rabits (approved):
Looks great!

PR #289, opened 2026-04-12
Update pre-commit hooks, move shared transform functions to transforms.py

GnoX:
This PR updates pre-commit hooks and consolidates shared transform functions into a dedicated module.

Full list of changes:
- Updated pre-commit hooks configuration and apply resulting formatting/lint fixes across affected files
- Moved `out_power`, `flow_is_on`, `pround`, `pmultiply`, `prop_has_bit_on`, `prop_has_bit_off` from `props/utils.py` into the new `props/transforms.py` module and updated all import sites to import from `transforms`
- Removed redundant `@dataclasses.dataclass` from `toggle` (already applied by `EntityType.__init_subclass__`)
- Fixed `NumberType.__call__` to return the limit-checking wrapper instead of the unwrapped function
- Removed duplicate `availability_prop` method from number builder (identical to parent class implementation)


GnoX:
## Code Review

### Overview
Three categories of changes:
1. **Pre-commit hook updates** — bumps ruff (v0.14→v0.15), pre-commit-hooks (v5→v6), swaps markdownlint for rumdl, adds pyproject-fmt and detect-private-key/check-added-large-files hooks
2. **Transform consolidation** — moves `out_power`, `flow_is_on`, `pround`, `pmultiply`, `prop_has_bit_on/off` into a shared `transforms.py`, removing duplicates from `_delta3_base`, `river3`, `delta_pro_3`, `delta3_max_plus`, and `utils.py`
3. **Controls cleanup** — removes redundant `@dataclasses.dataclass`, fixes `NumberType.__call__` return, removes duplicate `availability_prop`, simplifies `devicebase.py`

### Looks good
- Transform deduplication is clean — identical functions existed in 4 files, now single source of truth
- All import sites updated consistently, no missed references
- Dead `_out_power` in `delta3_max_plus` removed
- `devicebase.py` simplifications (`self.__doc__ or ""`, method reference for `connection_log.append`) are nice
- `NumberType.__call__` fix is correct — was returning the original `func` instead of the `_check_limits` wrapper, meaning limit checking was silently skipped
- `pyproject.toml` reformatted by pyproject-fmt — deps sorted, normalized names (`bleak-retry-connector` vs `bleak_retry_connector`), classifiers added

### Suggestions

- **`flow_is_on` type hint** — now that it's a shared utility, the parameter `x: int` is good, but the function calls `int(x)` which suggests it sometimes receives non-int values (floats from protobuf). Consider `x: int | float` to be accurate about actual callers.

- **`out_power` return type inconsistency** — `out_power(0.0)` returns `0` (int literal) while `out_power(1.0)` returns `-1.0` (float). Pre-existing, but now that it's promoted to shared code, could be worth normalizing to always return float: `return -round(x, 2) if x != 0 else 0.0`

- **Typo in PR description** — "upated" → "updated"

### Verdict
Clean PR. No blocking issues. Ship it.

rabits (approved):
Yep, cleaner now

PR #293, opened 2026-04-18
Reload config entry when offline BLE device reappears via advertisement

GnoX:
When a configured device isn't advertising at setup time, Home Assistant raises `ConfigEntryNotReady` and falls back to its fixed retry schedule which can take minutes or hours before the integration tries again. This PR makes the integration react immediately to BLE advertisements instead of waiting. This should help with #230, but also improve experience with other devices (https://github.com/GnoX/ha-ef-ble/discussions/74) and increase perceived stability.


rabits (approved):
Thank you, I hope this will help

PR #296, opened 2026-04-24
Fix STREAM Ultra PV2 sensor, consolidate rounding/division transforms and update pre-commit hooks

### LINKED ISSUE #294: [Bug]: PV(2) Power sensor has gone missing
### What happened?

PV1 and PV2 both have panels attached.  PV2 is generating OK (according to the PV total sensor, the EF app, and the mqtt feed from EF), but the sensor in HA went missing some days ago, after working fine for months.  
Restarting HA does not bring it back, and all other sensors/controls seem fine.    
How should I go about fixing/investigating this?

0.8.4 running on HA core

Stream Ultra X

Internal bluetooth adapter

_No response_

```text
no changes have been made (I can't seem to select No above!)
```

### Relevant log output

```shell
I've seen no relevant log entries
```

GnoX:
This PR fixes swapped PV2 and PV4 sensors for STREAM Ultra and does few clean ups. Full list of changes:

- **Fixed PV2 power sensor on STREAM Ultra**  - addresses #294 by swapping PV2 and PV4 sensors which were swapped because STREAM Max (with only 2 PV inputs) does use PV1 and PV4 internally, but for Max, we want to display it as PV1 and PV2 so we previously swapped it thinking that it is also swapped on other variants, but this was not true so this PR swaps it back
- **Consolidated rounding / scaling transforms** - added `pdiv(divisor, precision)` to `eflib/props/transforms.py` and replaced ad-hoc helpers and inline lambda patterns across all device modules having this patterns - no behavioural changes - the shared transforms are `None`-safe, matching existing `pround` semantics
- **Switched CI from `pre-commit` to `prek`** - `validate-code-style.yaml` now runs `uvx prek run --all-files` via `astral-sh/setup-uv` instead of `pre-commit/action` plus explicit Python setup - `prek` is a faster rust implementation of `pre-commit`

### DIFF
--- a/.github/workflows/validate-code-style.yaml
+++ b/.github/workflows/validate-code-style.yaml
@@ -10,7 +10,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.x"
-      - uses: pre-commit/action@v3.0.1
+      - uses: astral-sh/setup-uv@v5
+      - run: uvx prek run --all-files --show-diff-on-failure --color=always
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: check-added-large-files
         args: ['--maxkb=500']
   - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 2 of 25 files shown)


rabits (approved):
Thank you, it looks very nice!

PR #302, opened 2026-04-26
Fix plug index error

sidjohn1:
Fix plug index error when adjusting amps on secondary dc input plug on D3+

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta3_plus.py
+++ b/custom_components/ef_ble/eflib/devices/delta3_plus.py
@@ -30,7 +30,15 @@ class Device(delta3.Device):
         return 0
 
     @controls.current(dc_charging_max_amps_2, max=dynamic(dc_charging_current_max_2))
-    async def set_dc_charging_amps_max_2(self, value: float) -> bool:
-        return await self.set_dc_charging_amps_max(
-            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_2
-        )
+    async def set_dc_charging_amps_max_2(
+        self,
+        value: float,
+        plug_index: pd335_sys_pb2.PV_PLUG_INDEX = pd335_sys_pb2.PV_PLUG_INDEX_2,
+    ) -> bool:
+        config = pd335_sys_pb2.ConfigWrite()
+        config.cfg_pv_dc_chg_setting.pv_plug_index = plug_index
+        config.cfg_pv_dc_chg_setting.pv_chg_vol_spec = pd335_sys_pb2.PV_CHG_VOL_SPEC_12V
+        config.cfg_pv_dc_chg_setting.pv_chg_amp_limit = int(value)
+
+        await self._send_config_packet(config)
+        return True


custom_components/ef_ble/eflib/devices/delta3_plus.py
```diff
@@ -30,7 +30,15 @@ def solar_input_power_2(self) -> float:
         return 0
 
     @controls.current(dc_charging_max_amps_2, max=dynamic(dc_charging_current_max_2))
-    async def set_dc_charging_amps_max_2(self, value: float) -> bool:
-        return await self.set_dc_charging_amps_max(
-            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_2
-        )
```
GnoX, line 36: Ah, right, I screwed up by not allowing other args in limit checker decorator. Instead of doing this redefinition, can you go into `eflib/entity/controls.py` and replace inner function on line 76 with this?
```python
        async def _check_limits(
            device: "DeviceBase", value: float, *args, **kwargs
        ) -> bool:
            if (low := _resolve(control.min, device)) is not None:
                value = max(low, value)
            if (high := _resolve(control.max, device)) is not None:
                value = min(high, value)
            return await func(device, value, *args, **kwargs)
```

sidjohn1: i can give this a shot later tonight, but if we revert lines 34 - 36 in custom_components/ef_ble/eflib/devices/delta3_plus.py and put the fix in eflib/entity/controls.py. shouldn't lines 192 - 203 in custom_components/ef_ble/eflib/devices/_delta3_base.py be like lines 34 - 36 in delta3_plus.py.

Please keep in mind i'm not a programmer, i'm a good pattern matcher. So, in my head shouldn't the code for PV_PLUG_INDEX_1 be very similar to the code for PV_PLUG_INDEX_2?

GnoX: This code worked until I decided to implement that @control system - `delta3_plus.py` uses that function, it just passes `PV_PLUG_INDEX_2` there as an argument. 

Of course, you can do what you did here - it works, but introduces unnecessary duplication. It's generally better to reuse rather than copy.

sidjohn1: I may have explained myself poorly. hopefully this makes more sense.

If it makes more sense to change eflib/entity/controls.py with the following, then i'm on board with giving this a shot.
```python
         async def _check_limits(
            device: "DeviceBase", value: float, *args, **kwargs
        ) -> bool:
            if (low := _resolve(control.min, device)) is not None:
                value = max(low, value)
            if (high := _resolve(control.max, device)) is not None:
                value = min(high, value)
            return await func(device, value, *args, **kwargs)
```
            
 In agreement with you on limiting unnecessary duplication, if this good in custom_components/ef_ble/eflib/devices/delta3_plus.py
 
 ```python
    @controls.current(dc_charging_max_amps_2, max=dynamic(dc_charging_current_max_2))
    async def set_dc_charging_amps_max_2(self, value: float) -> bool:
        return await self.set_dc_charging_amps_max(
            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_2
        )
```
        
 wouldn't we also want to make this change in custom_components/ef_ble/eflib/devices/_delta3_base.py 
 
 ```python
    @controls.current(dc_charging_max_amps, max=dynamic(dc_charging_current_max))
    async def set_dc_charging_amps_max(
        self,
        value: float,
        plug_index: pd335_sys_pb2.PV_PLUG_INDEX = pd335_sys_pb2.PV_PLUG_INDEX_1,
    ) -> bool:
        config = pd335_sys_pb2.ConfigWrite()
        config.cfg_pv_dc_chg_setting.pv_plug_index = plug_index
        config.cfg_pv_dc_chg_setting.pv_chg_vol_spec = pd335_sys_pb2.PV_CHG_VOL_SPEC_12V
        config.cfg_pv_dc_chg_setting.pv_chg_amp_limit = int(value)
        await self._send_config_packet(config)
        return True
```
        
to this?
        
```python
    @controls.current(dc_charging_max_amps_1, max=dynamic(dc_charging_current_max_1))
    async def set_dc_charging_amps_max_1(self, value: float) -> bool:
        return await self.set_dc_charging_amps_max(
            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_1
        )
```

If custom_components/ef_ble/eflib/devices/_delta3_base.py needs to be left alone to function normally you can just say that.

GnoX: I see, you're talking about naming semantics, right? If I understood correctly, the issue you see is the assymetric naming - main function does not have `_1` suffix. This has a reason - regular `Delta 3` does only have one of those sliders and it makes more sense to just call it `DC Charging Max Amps` rather than `DC (1) Charging Max Amps` because there is not and won't be a second one. We could change it and define both 1 and 2 for D3+ but it's getting more complicated in this case so I this was the easiest way to do this.

sidjohn1: no sorry, more like this
```python
    @controls.current(dc_charging_max_amps, max=dynamic(dc_charging_current_max))
    async def set_dc_charging_amps_max(self, value: float) -> bool:
        return await self.set_dc_charging_amps_max(
            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_1
        )
```
        
but i may just need to sit this out and learn from how you fix this.

GnoX: Well, that would be okay, but you would have to rename that function you're calling because it would just be calling itself.

This would be a valid fix:
```python
   async def _set_dc_charging_amps_max(
       self,
       value: float,
       plug_index: pd335_sys_pb2.PV_PLUG_INDEX,
   ) -> bool:
       config = pd335_sys_pb2.ConfigWrite()
       config.cfg_pv_dc_chg_setting.pv_plug_index = plug_index
       config.cfg_pv_dc_chg_setting.pv_chg_vol_spec = pd335_sys_pb2.PV_CHG_VOL_SPEC_12V
       config.cfg_pv_dc_chg_setting.pv_chg_amp_limit = int(value)
       await self._send_config_packet(config)
       return True

    @controls.current(dc_charging_max_amps, max=dynamic(dc_charging_current_max))
    async def set_dc_charging_amps_max(self, value: float) -> bool:
        return await self._set_dc_charging_amps_max(
            value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_1
        )

   # in delta3_plus.py
   @controls.current(dc_charging_max_amps_2, max=dynamic(dc_charging_current_max_2))
   async def set_dc_charging_amps_max_2(self, value: float) -> bool:
       return await self._set_dc_charging_amps_max(
           value, plug_index=pd335_sys_pb2.PV_PLUG_INDEX_2
       )
```

GnoX:
Fixed in #310

PR #306, opened 2026-05-01
Update delta2_max.py charge speed

### LINKED ISSUE #305: number.ef_d2max_blu_ac_charging_speed - limits for D2Max are not correct
### Pre-submission checks

### Device model

D2Max

### Sensor / control name

number.ef_d2max_blu_ac_charging_speed

This number set valuenumber.ef_d2max_blu_ac_charging_speed seems to limited to 1800 W - that's incorrect - max charge speed for D2Max is 2400 W

### Integration version

0.8.5

### Diagnostics

_No response_

### Anything else?

_No response_

florismetzner:
specheet says 2400 W max charge speed (combines AC/DC) https://www.pollin.at/media/ea/ef/c5/1701705432/D273131-D.pdf

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta2_max.py
+++ b/custom_components/ef_ble/eflib/devices/delta2_max.py
@@ -26,7 +26,7 @@ class Device(Delta2Base):
 
     @computed_field
     def max_ac_charging_power(self) -> int:
-        return 1800
+        return 2400
 
     @property
     def pd_heart_type(self):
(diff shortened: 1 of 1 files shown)


custom_components/ef_ble/eflib/devices/delta2_max.py
```diff
@@ -26,7 +26,7 @@ class Device(Delta2Base):
 
     @computed_field
     def max_ac_charging_power(self) -> int:
-        return 1800
+        return 2400
```
GnoX: Actually, there is a field that tells the max value, I overlooked it, if you add a field
```python
    ac_chg_rated_power = raw_field(pb_inv.ac_chg_rated_power)
```

Then this can be

```suggestion
        return self.ac_chg_rated_power or 1800
```

florismetzner (commented):
Looks good

custom_components/ef_ble/eflib/devices/delta2_max.py
florismetzner: Done, implemented your suggestion

GnoX (approved):

rabits (approved):
Nice & clean! Good.

PR #307, opened 2026-05-01
Add V4 wire format support, refactor Packet to dataclasses and align naming scheme for packets

GnoX:
This PR adds support for the V4 wire format used by newer EcoFlow devices (e.g. SHP3, DPU X) and refactors the existing `Packet` and `EncPacket` into a dataclasses and aligns naming of their attributes to snake case.

Full list of changes:

- **Added `PacketV4` codec for the V4 wire format** - implements the two-layer XOR obfuscation (CRC8 key over the inner cmd block plus optional `v4_type_b` key over the application payload) and a CRC16 trailer computed over the obfuscated bytes - V4 packets are used
- **Transparent dispatch from `Packet.from_bytes`** - V3 and V4 frames go through the same entrypoint; when `version == 0x04` parsing is delegated to `PacketV4.from_bytes` so device modules don't need to know which wire format they received
- **Refactored `Packet`/`PacketV4` to `@dataclass`** - replaces the hand-written `__init__`/property soup with a dataclass and a few `cached_property` derived fields, and aligns naming to snake_case (`cmd_set`/`cmd_id`/`payload_hex`/`product_id` instead of mixed `cmdSet`/`cmdId`/`payloadHex`/`productId`); call sites across all device modules updated accordingly - no behavioural changes

### DIFF
--- a/custom_components/ef_ble/eflib/packet.py
+++ b/custom_components/ef_ble/eflib/packet.py
@@ -1,5 +1,6 @@
 import logging
 import struct
+from functools import cached_property
 from typing import TypeGuard
 
 from .crc import crc8, crc16
@@ -8,8 +9,6 @@ _LOGGER = logging.getLogger(__name__)
 
 
 class Packet:
-    """Needed to parse and make the internal packet structure"""
-
     PREFIX = b"\xaa"
 
     NET_BLE_COMMAND_CMD_CHECK_RET_TIME = 0x53
@@ -39,9 +38,6 @@ class Packet:
         self._seq = seq if seq is not None else b"\x00\x00\x00\x00"
         self._product_id = product_id
 
-        # For representation
-        self._payload_hex = bytearray(self._payload).hex()
-
     @property
     def src(self):
         return self._src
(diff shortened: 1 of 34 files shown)


rabits:
Thank you, will check it and the others this weekend.

custom_components/ef_ble/eflib/packet.py
```diff
@@ -1,110 +1,65 @@
...
-        self._dsrc = dsrc
-        self._ddst = ddst
-        self._version = version
-        self._seq = seq if seq is not None else b"\x00\x00\x00\x00"
-        self._product_id = product_id
-
-        # For representation
-        self._payload_hex = bytearray(self._payload).hex()
+    """
+    V2 / V3 / V0x13 packet codec
```
rabits: A bit of misalignment - probably you've meant V19 here (according to the docs in frame_assembler.py)

rabits (approved):
Overall is good, just a little thing with V19/v0x13

custom_components/ef_ble/eflib/packet.py
GnoX: Thanks, that's another one of those silly little AI artifacts.

PR #310, opened 2026-05-03
Fix energy backup and backup reserve not working on latest Delta 3 firmwares, fix D3+ DC (2) Max Amp slider

### LINKED ISSUE #309: [Sensor]: backup reserve
### Pre-submission checks

### Device model

Delta 3, firmware: v7.73.86.33

### Sensor / control name

Backup reserve 

Bonjour, 

Image

Pouvez-vous regarder mon problème de capteur backup reserve il fonctionnait avant mise à jour du firmware v7.73.86.33 ancien : v7.73.86.19
Ça m’embête car je m’en sert pour recharger ma delta quand production solaire suffisante..
Merci par avance 

### Integration version

0.8.5

### Diagnostics

_No response_

### Anything else?

_No response_

GnoX:
This PR fixes the energy backup switch and backup reserve level on all Delta 3 models against the latest device firmware, where the protobuf fields driving these controls are no longer present - we were using fields used in River 3 which may have been included in Delta 3 firmware, but that wasn't the right way to do it as app does not recognize `Energy Backup` as a control - it is a `Self Powered Mode`.

PR also fixes second Delta 3 Plus DC slider that was broken when porting it to controls - limit checking decorator was not passing all args to the decorated method. Supersedes #302

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta3_classic.py
+++ b/custom_components/ef_ble/eflib/devices/delta3_classic.py
@@ -1,7 +1,7 @@
 from ..entity import controls
 from ..entity.base import dynamic
 from ..pb import pd335_sys_pb2
-from ..props import pb_field
+from ..props import computed_field, pb_field
 from ..props.transforms import flow_is_on, out_power
 from ._delta3_base import Delta3Base, pb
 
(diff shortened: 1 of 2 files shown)


rabits (approved):
Looks good!

PR #319, opened 2026-05-10
Close PR #319 - will revisit in a cleaner follow-up

mariusangelmann:
Closing this PR for now.

The branch needs cleanup and I don't want to keep an unclear diff open for review. I'll come back with a smaller, cleaner PR if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,27 +358,30 @@ Click on any device below to see available sensors, switches, and controls:
 
 <br>
 
-| *Sensors*                           | *Switches*         | *Sliders*                 | *Selects*       |
-|-------------------------------------|--------------------|---------------------------|-----------------|
-| Battery Level                       | Feed Grid          | Feed Grid Power Limit     | Energy Strategy |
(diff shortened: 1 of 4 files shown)


GnoX:
This is a lot of code for something that can be solved in few clicks and you can't even configure. This is just going to make it super hard to implement my upcoming rework of sensors so if we decide to go this way, keep in mind that this will definitely be broken. 

It also does not follow the sensor registration style, unnecessarily reimplements integration while it can just use `IntegrationSensor` from HA that does all of this with a lot less code, reimplements added/removed callbacks incorrectly, there is a bug where all of the keys are the same causing unique id conflicts, I think that it silently zeroes corrupted state without logging anything, the test is super useless and other issues that I noticed just by quickly skimming through it. 

I cannot in good conscience approve this, in fact we're going to close this as per [contribution guidelines](https://github.com/rabits/ha-ef-ble?tab=contributing-ov-file#ai-assisted-contributions). I don't really want to bother going through full review because it does look like it's AI generated and I'd rather reimplement it myself than waste time arguing with AI.

PR #324, opened 2026-05-12
Fix entity callback registration when device lacks the property

### LINKED ISSUE #323: [Bug]: River 2 Max AC ports switch unavailable and not working
### Pre-submission checks

### What happened?

After updating the integration, AC ports control stopped working for River 2 Max.

Current behavior:
- switch.river2max_ac_ports is always "unavailable"
- AC output is active and working (sensor.river2max_ac_output_power shows load ~75W)
- Other entities work correctly (DC port switch, sensors, telemetry)
- Automations using switch.turn_off no longer work

Previously:
- switch.river2max_ac_ports worked correctly and allowed enabling/disabling AC output

1. Update ha-ef-ble integration to latest version
2. Check entity switch.river2max_ac_ports
3. Try to toggle AC ports via UI or service call

Result:
- Entity is unavailable
- Control does not work

switch.river2max_ac_ports should:
- reflect current AC state
- allow turning AC ports on/off

### Integration version

0.8.7

### Home Assistant version

2026.5.1

### Device model

River 2 Max

### Bluetooth connection

Internal bluetooth adapter

No

```text

```

### Diagnostics

_No response_

### Relevant log output

```shell

```

GnoX:
This PR fixes the AC ports switch on River 2 Max (and likely other unmigrated devices) being stuck as unavailable since 0.8.6. Fix is generic rather than tweaking the deprecated entry: _register_update_callback now bails out when the device doesn't expose prop_name, so any future description referencing a state or availability prop the device doesn't implement is silently ignored instead of bricking the entity.

### DIFF
--- a/custom_components/ef_ble/entity.py
+++ b/custom_components/ef_ble/entity.py
@@ -49,7 +49,7 @@ class EcoflowEntity(Entity):
         get_state: Callable[[Any], SkipWrite | Any] = lambda x: x,
         default_state: Any = None,
     ):
-        if prop_name is None:
+        if prop_name is None or not hasattr(self._device, prop_name):
             return
 
         if value := getattr(self._device, prop_name, None):


rabits (approved):
Looks good!

PR #325, opened 2026-05-14
Save diagnostics to disk on connection error

GnoX:
This PR adds an opt-in option to automatically persist device diagnostics to disk when a BLE connection enters an error state, making it easier to troubleshoot intermittent connection / auth failures after the fact.

- New `with_diagnostics_on_exception(enabled)` builder on `DeviceBase` and matching `diagnostics_on_exception` toggle in the diagnostics options of the config flow. When enabled, packet collection is force-enabled and a connection-state listener saves the collected diagnostics to disk on any error state.
- Captures sent raw data and the initial session key in the diagnostics payload so post-mortem analysis has the full session context, not just received frames.
- Refactors `diagnostics.py` to build the dict via `DeviceDiagnosticsCollector.build_diagnostics_dict()`, consolidating the per-device fields in one place.

### DIFF
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -29,6 +29,7 @@ from .const import (
     CONF_BLUEZ_START_NOTIFY,
     CONF_COLLECT_PACKETS_AMOUNT,
     CONF_CONNECTION_TIMEOUT,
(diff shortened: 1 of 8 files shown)


rabits (approved):
Yep, looks good!

PR #326, opened 2026-05-16
Add initial support for Delta Pro Ultra X, type 0 encryption and heartbeat-based auth skip

### LINKED ISSUE #102: Delta Pro Ultra X support
I am trying to access my new Ecoflow through this integration. Steps taken:

1. The ESP32 bluetooth proxy is up an running. Used the config settings posted in the Wiki. In the "Advertisement Monitor" I can see the Ecoflow bluetooth - EF-P1020191. Have also confirmed that the app is not using bluetooth on my phone. 
2. Loaded the custom integration - EcoFlow BLE (Unofficial) - into Home Assistant. 

When I click to add the Ecoflow BLE integration, I immediately receive the following error: "No supported EcoFlow devices found. Make sure your device is not connected to the EcoFlow app via Bluetooth." 

It provides no other options to manually search or anything else. I doesn't seem to be a bluetooth issue as I see the Ecoflow in the ESP32 proxy. 

Any thoughts on what I am missing? Thanks for the help.

GnoX:
This PR adds first-pass support for the EcoFlow Delta Pro Ultra X (`EF-P10*`, SN prefix `P101`) on top of the V4 wire format from #307, and bundles two connection-layer changes the DPU X needs to authenticate at all: a sentinel-format / unencrypted variant of the V2/V3 framing ("type 0 encryption"), and a short-circuit in the auth wait when the device starts pushing heartbeat frames before the integration's auth state machine has run.

Full list of changes:

- **Added type 0 (unencrypted) packet framing** - introduces a sentinel-format variant of the V2/V3 framing where the high-nibble bit on the version byte signals a `0xBB 0xBB` payload sentinel in lieu of a trailing CRC16. Wired through a new `PassthroughAssembler` that the connection layer selects when the device negotiates `encrypt_type=0`. Required for the DPU X's auth and key-exchange phase, which is sent unencrypted.
- **Skip waiting for auth if the device already sent a heartbeat** - DPU X starts emitting heartbeat frames as soon as the BLE link is up, before the integration's auth state machine has run. `listenForDataHandler` now treats the first non-auth data packet during `AUTHENTICATING` as evidence that the device considers the session authenticated and transitions to `AUTHENTICATED` instead of timing out the auth wait.
- **Added initial support for Delta Pro Ultra X device** - new `eflib/devices/dpu_x.py` (protobuf-based via `pd100_pb2`), `proto/pd100.proto` definitions, and the `EF-P10*`
- **Fixed V4 frame size accounting in `PassthroughAssembler`** - the V4 frame-size field was being read with the wrong width, causing the assembler to drop trailing bytes on longer DPU X frames.

Resolves #102


rabits (approved):
Very nice!

PR #335, opened 2026-05-24
Add CN region support and phone number login for EcoFlow China users

maybetaken:
Changes:
- Add 'CN' region option (api-cn.ecoflow.com)
- Support phone number login with 'phone' field
- Auto-detect phone numbers and route to CN region
- Strip +86 prefix if present in phone number
- Update translations for Email/Phone field label

### DIFF
--- a/custom_components/ef_ble/config_flow.py
+++ b/custom_components/ef_ble/config_flow.py
@@ -513,26 +513,41 @@ class EFBLEConfigFlow(ConfigFlow, domain=DOMAIN):
     async def _ecoflow_login(self, email: str, password: str, region: str):
         session = async_get_clientsession(self.hass)
 
+        identifier = email.strip()
+        digits = identifier.removeprefix("+").replace(" ", "")
+        is_phone = digits.isdigit() and 6 <= len(digits) <= 15
+
         match region:
             case "EU":
                 base_url = "api-e.ecoflow.com"
             case "US":
                 base_url = "api-a.ecoflow.com"
+            case "CN":
+                base_url = "api-cn.ecoflow.com"
             case _:
-                base_url = "api.ecoflow.com"
+                # Auto-detect: phone numbers default to China region
+                base_url = "api-cn.ecoflow.com" if is_phone else "api.ecoflow.com"
+
+        json_payload = {
+            "scene": "IOT_APP",
+            "appVersion": "1.0.0",
+            "password": base64.b64encode(password.encode()).decode(),
+            "oauth": {
+                "bundleId": "com.ef.EcoFlow",
+            },
+            "userType": "ECOFLOW",
+        }
+
+        if region == "CN" and not is_phone:
+            return {"login": "CN region requires phone number, not email"}
+        if region not in ("EU", "US") and is_phone:
+            json_payload["phone"] = identifier.removeprefix("+86")
+        else:
+            json_payload["email"] = identifier
 
         async with session.post(
             url=f"https://{base_url}/auth/login",
-            json={
-                "scene": "IOT_APP",
-                "appVersion": "1.0.0",
-                "password": base64.b64encode(password.encode()).decode(),
-                "oauth": {
-                    "bundleId": "com.ef.EcoFlow",
-                },
-                "email": email,
-                "userType": "ECOFLOW",
-            },
+            json=json_payload,
             headers={
                 "Accept": "application/json",
                 "Content-Type": "application/json",
@@ -675,7 +690,9 @@ class _SchemaBuilder:
                         schema_builder()
                         .optional(CONF_EMAIL, str)
                         .optional(CONF_PASSWORD, str)
-                        .optional(CONF_REGION, vol.In(["Auto", "EU", "US"]), "Auto")
+                        .optional(
+                            CONF_REGION, vol.In(["Auto", "EU", "US", "CN"]), "Auto"
+                        )
                         .build()
                     ),
                     options={"collapsed": collapsed},
--- a/custom_components/ef_ble/translations/en.json
+++ b/custom_components/ef_ble/translations/en.json
@@ -43,12 +43,12 @@
             "name": "EcoFlow Login (skip if you already have your user ID)",
             "description": "After you click Submit with your login information filled in, your EcoFlow User ID will be automatically populated and this form will clear itself. Only your user ID is required so you will not be able to proceed to the next page unless this section is empty.",
             "data": {
-              "email": "Email",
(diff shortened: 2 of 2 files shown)


custom_components/ef_ble/config_flow.py
```diff
@@ -513,26 +513,44 @@ def _store(self):
     async def _ecoflow_login(self, email: str, password: str, region: str):
         session = async_get_clientsession(self.hass)
 
+        is_phone = "@" not in email
```
GnoX: We should make this detection a little bit more robust, something like this:

```suggestion
        identifier = email.strip()
        digits = identifier.removeprefix("+").replace(" ", "")
        is_phone = digits.isdigit() and 6 <= len(digits) <= 15
```

custom_components/ef_ble/config_flow.py
```diff
@@ -513,26 +513,44 @@ def _store(self):
...
                 base_url = "api-a.ecoflow.com"
+            case "CN":
+                base_url = "api-cn.ecoflow.com"
             case _:
-                base_url = "api.ecoflow.com"
+                # Auto-detect: phone numbers default to China region
+                if is_phone:
+                    base_url = "api-cn.ecoflow.com"
+                else:
+                    base_url = "api.ecoflow.com"
```
GnoX: base_url = "api-cn.ecoflow.com" if is_phone else "api.ecoflow.com"

custom_components/ef_ble/config_flow.py
```diff
@@ -513,26 +513,44 @@ def _store(self):
...
+        }
+
+        if region == "CN" or (region not in ("EU", "US") and is_phone):
+            if not is_phone:
+                return {"login": "CN region requires phone number, not email"}
+            # Strip +86 prefix if present
+            phone = email.removeprefix("+86") if email.startswith("+86") else email
+            json_payload["phone"] = phone
+        else:
+            json_payload["email"] = email
```
GnoX: I would do this whole thing as


```suggestion
        if region == "CN" and not is_phone:
            return {"login": "CN region requires phone number, not email"}

        if region not in ("EU", "US") and is_phone:
            json_payload["phone"] = identifier.removeprefix("+86")
        else:
            json_payload["email"] = identifier
```

GnoX (changes requested):

custom_components/ef_ble/config_flow.py
maybetaken: Done

custom_components/ef_ble/config_flow.py
maybetaken: Done

custom_components/ef_ble/config_flow.py
maybetaken: LGTM, Done.

GnoX (approved):
Looks good, thanks. I wasn't aware that CN region even existed - when switching to China, it showed `api-a` in the documentation, but I didn't realize that you have to go to entirely different developer portal that only shows China.

BTW: I also added CN region to https://gnox.github.io/user_id

maybetaken:
> Looks good, thanks. I wasn't aware that CN region even existed - when switching to China, it showed `api-a` in the documentation, but I didn't realize that you have to go to entirely different developer portal that only shows China.
> 
> BTW: I also added CN region to https://gnox.github.io/user_id

Thank you, and I deem you should also add this link in the front page so that all users could find it easily.

maybetaken:
> Looks good, thanks. I wasn't aware that CN region even existed - when switching to China, it showed `api-a` in the documentation, but I didn't realize that you have to go to entirely different developer portal that only shows China.
> 
> BTW: I also added CN region to https://gnox.github.io/user_id

I test this web, as when I use phone number and `AUTO` option, it didn't use CN api, could this be improved?

GnoX:
> Thank you, and I deem you should also add this link in the front page so that all users could find it easily.

We do have it linked in our wiki, I'll think about putting it in the README, don't really want to refer people to external tools from there.

> I test this web, as when I use phone number and AUTO option, it didn't use CN api, could this be improved?

Thanks for the feedback, it should now route to CN properly.

Also, I noticed that I had the regions wrong as I always thought that `api-a` meant America as opposed to Asia. This means that our config flow here is also wrong, it should be like this:
- US - Global / United States (api.ecoflow.com)
- APAC - Asia-Pacific (api-a.ecoflow.com)
- EU - Europe (api-e.ecoflow.com)
- CN - China (api-cn.ecoflow.com)

`api.ecoflow.com` does not route to region automatically so there is no `Auto` mode, this is what I was missing when people were having problems with their user ids.

I'll fix it in the next PR or if you feel like it, you can fix it here.

maybetaken:
> > Thank you, and I deem you should also add this link in the front page so that all users could find it easily.
> 
> We do have it linked in our wiki, I'll think about putting it in the README, don't really want to refer people to external tools from there.
> 
> > I test this web, as when I use phone number and AUTO option, it didn't use CN api, could this be improved?
> 
> Thanks for the feedback, it should now route to CN properly.
> 
> Also, I noticed that I had the regions wrong as I always thought that `api-a` meant America as opposed to Asia. This means that our config flow here is also wrong, it should be like this:
> 
> * US - Global / United States (api.ecoflow.com)
> * APAC - Asia-Pacific (api-a.ecoflow.com)
> * EU - Europe (api-e.ecoflow.com)
> * CN - China (api-cn.ecoflow.com)
> 
> `api.ecoflow.com` does not route to region automatically so there is no `Auto` mode, this is what I was missing when people were having problems with their user ids.
> 
> I'll fix it in the next PR or if you feel like it, you can fix it here.

Maybe next PR, I am in the middle of translating it into Chinese.

GnoX:
No worries, I'll do it myself.

> I am in the middle of translating

Translations are a mess right now, there's so many duplicatates because we can't use `strings.json` from custom integrations. I do have something planned for this, but I didn't get to it yet.

maybetaken:
> Translations are a mess right now, there's so many duplicatates because we can't use `strings.json` from custom integrations. I do have something planned for this, but I didn't get to it yet.

It looks ok to me, as this is the only we could do that... BTW, could PR merged?

GnoX:
> It looks ok to me, as this is the only we could do that...

Well, we could have a script that compiles it similar to how HA does it and have CI job that checks it.

> BTW, could PR merged?

Merging is always up to @rabits

maybetaken:
I see

rabits (approved):
Yeah looks nice, thank you for contribution!

PR #336, opened 2026-05-24
Fix region mapping for EF login from config flow, refactor login into a separate class

GnoX:
This PR consolidates region routing and credential probing that lived inline in the config flow into a dedicated `eflib.login` helper, fixes the region map that had wrong labels.
PR also adds an automatic detection of region by trying regions in order of US/Global, EU, APAC, though this may sometimes yield incorrect results.

### DIFF
--- a/custom_components/ef_ble/eflib/login.py
+++ b/custom_components/ef_ble/eflib/login.py
@@ -1,0 +1,144 @@
+"""EcoFlow account login helper - region routing and credential probing."""
+
+import base64
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import Any
+
+import aiohttp
+
+
+class Region(StrEnum):
+    """Selectable EcoFlow account region"""
+
+    AUTO = "Auto"
+    US = "US"
+    EU = "EU"
+    APAC = "APAC"
+    CN = "CN"
+
+    @classmethod
+    def _missing_(cls, value: object) -> "Region | None":
+        if isinstance(value, str):
+            for member in cls:
+                if member.value.lower() == value.lower():
+                    return member
+        return None
+
+    @property
+    def base_url(self) -> str | None:
+        """API host for this region, or None for AUTO."""
+        return _REGION_BASE_URL.get(self)
+
+
+_REGION_BASE_URL: dict[Region, str] = {
+    Region.US: "api.ecoflow.com",
+    Region.EU: "api-e.ecoflow.com",
+    Region.APAC: "api-a.ecoflow.com",
+    Region.CN: "api-cn.ecoflow.com",
+}
+
+# AUTO tries these in order for email logins. Phone numbers go straight to CN.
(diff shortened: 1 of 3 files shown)


custom_components/ef_ble/eflib/login.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def base_url(self) -> str | None:
+        """API host for this region, or None for AUTO."""
+        return _REGION_BASE_URL.get(self)
+
+
+_REGION_BASE_URL: dict[Region, str] = {
+    Region.US: "api.ecoflow.com",
+    Region.EU: "api-e.ecoflow.com",
+    Region.APAC: "api-a.ecoflow.com",
```
maybetaken: It should be the server that user connected, not specific country, I think. AFAIK, HK CHINA uses US api-a, which is totally different from China mainland. For US, the api website of ecoflow describes it use api-a, not api, but it also works for US.

GnoX: Ah, right, thanks. I had it right the first time, it seems that America does use `api-a`, but it also includes Australia and Oceania while `api-e` is used by European, Asian and African countries. `api.ecoflow.com` is attempting to resolve different servers depending on the proxy I use. Will revert this back to what we had before.

GnoX: Scratch that proxy resolving, its just hitting a traffic manager service.

Did some scraping from the developer web and found this:
```
+-----------------------+-----------+----------------------------------------+
| host                  | countries | members                                |
+-----------------------+-----------+----------------------------------------+
| api.ecoflow.com       |    57     | US, CA, MX, BR, AR, AU, PH, TW, LatAm  |
| api-e.ecoflow.com     |    86     | EU/EEA, UK, KR, NZ, ZA, JM             |
| api-a.ecoflow.com     |   103     | IN, HK, MO, SG, ID, VN, SA, AE, TR, MEA|
| api-j.ecoflow.com     |     1     | JP                                     |
| api-r.ecoflow.com     |     1     | RU                                     |
| api-cn.ecoflow.com    |    -      | China mainland (separate auth flow)    |
+-----------------------+-----------+----------------------------------------+
```

However, the countries seem to shift to different regions depending on the developer portal you use, these results are from `developer.ecoflow.com`, but if you visit `developer-eu.ecoflow.com`, most countries in Asia are in `api-e`.

GnoX: Changed it so that it lists API codes first in the selects and added a note that says that geo region may not correspond to the account region. d68ce46

rabits (approved):
Yeah, very nice!

custom_components/ef_ble/eflib/login.py
rabits: @maybetaken could you please resolve if it's not an issue anymore?

maybetaken: LGTM

PR #337, opened 2026-05-24
Fix PowerStream battery level over-reporting in upper SoC range

GnoX:
This PR fixes a long-standing discrepancy where the PowerStream's `Battery Level` sensor reads different values than the actual SoC in the upper range (GnoX/ha-ef-ble#79), visible when compared side-by-side with a connected device on the same battery via BLE and via the cloud.


rabits (approved):
Looks good!

PR #342, opened 2026-05-25
Add River 3 Plus support, battery health, add-on batteries, errors

shenki:
Developed with help of Claude Code


GnoX:
Does ANY of this even work?
1. You added serial number to the River 3 instead of River 3 Plus which already contains your prefix
2. Most of the sensors you added don't conform to our standard naming and are not registered in sensor.py
3. River 3 Plus does only support 1 additional battery, you added 2 of them and even has a wrong name

I doubt you even tested anything, closing in accordance with our [AI guidelines](https://github.com/rabits/ha-ef-ble?tab=contributing-ov-file#ai-assisted-contributions).

shenki:
Sure.

PR #349, opened 2026-05-28
Fix Delta Pro 3 AC Input Power sensor

### LINKED ISSUE #348: AC Input Power for DP3 not updating
Im using the latest version of Home Assistant and the app.
Im trying to track energy input to the Delta Pro 3.
There is an entity being provided (ac_input_power), but it never changes from 0.

This datapoint is crucial in understanding the activity around the unit.

Is the output broken?

GnoX:
We were using incorrect field for determining AC input power sensor - we didn't notice that it does not follow the same pattern as Delta 3 or River 3 where `pow_get_ac` has positive and negative values - in this case, it has separate field for AC input power - `pow_get_ac_in`.

Resolves #348

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta_pro_3.py
+++ b/custom_components/ef_ble/eflib/devices/delta_pro_3.py
@@ -41,7 +41,7 @@ class Device(DeviceBase, ProtobufProps):
     battery_level = pb_field(pb.cms_batt_soc, pround(2))
     battery_level_main = pb_field(pb.bms_batt_soc, pround(2))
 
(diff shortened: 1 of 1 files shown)


GnoX:
Unfortunately, pre-commit started failing because of https://github.com/tox-dev/toml-fmt/issues/355. I don't really want to really pin the `toml-fmt-common` just for pre-commit so we either have to wait for either compatibility alias or a new release of `pyproject-fmt`.

rabits (approved):
Small & simple) Yeah I think we can ignore this minor error of validator for now.

PR #356, opened 2026-05-29
Fix Delta 2 AC Ports switch stuck unavailable while AC output is off

### LINKED ISSUE #334: [Bug]: AC Ports switch on delta 3 1500 are sometimes "unavailable"
### Pre-submission checks

### What happened?

Sometimes the AC Ports switch becomes unavailable.
Usually after restarts or power outage. all other entities are working well. just this one is "unavailable" and this is very problematic as this is the switch needed to turn on locally during a blackout. the way to solve it is to physically go to the unit and turn it on, and then it becomes available again in the integration.
it's not consistent, sometimes a restart causes it and sometimes it happens by its own.

### Integration version

0.90

### Home Assistant version

2026.05.04

### Device model

Delta 3 1500

### Bluetooth connection

Internal bluetooth adapter

No

```text
Sometimes the AC Ports switch becomes unavailable.
Usually after restarts or power outage. all other entities are working well. just this one is "unavailable" and this is very problematic as this is the switch needed to turn on locally during a blackout. the way to solve it is to physically go to the unit and turn it on, and then it becomes available again in the integration.
```

### Diagnostics

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #351: Delta 2: switch ac_ports shows 'unavailable' instead of 'off' after reboot and cannot be turned on via HA
### Pre-submission checks

### What happened?

Title: Delta 2: switch ac_ports shows 'unavailable' instead of 'off' after reboot and cannot be turned on via HA

Device: EcoFlow Delta 2 (R331)
Integration version: 0.9.1
Home Assistant version: 2026.5.4
Installation type: Home Assistant Container (aarch64, Raspberry Pi)

## Description

After a Raspberry Pi reboot, switch.ef_r330429_ac_ports shows 'unavailable' 
instead of 'off'. The switch remains unavailable and cannot be controlled 
via Home Assistant until the AC port is turned on either by using the 
official EcoFlow mobile app or by pressing the physical AC button on the 
device itself. Once turned on manually, the switch becomes available and 
fully controllable via Home Assistant.

## Behavior

- After a Raspberry Pi reboot, switch.ef_r330429_ac_ports is 'unavailable'
- All other entities (sensors, switch.ef_r330429_usb_ports) are immediately available
- Calling switch.turn_on on the unavailable switch has no effect - 
  the BLE command is not sent and the AC port does not turn on physically
- The switch only becomes available and controllable after:
  - Turning on the AC port via the official EcoFlow mobile app, OR
  - Pressing the physical AC button on the Delta 2 unit
- After a Home Assistant restart (without Raspberry Pi reboot) the 
  switch remains in its last known state and works correctly

## Expected Behavior

switch.ef_r330429_ac_ports should show 'off' when the AC port is off 
and should respond to switch.turn_on commands regardless of the current state.

## Workaround

Turn on the AC port via the EcoFlow mobile app or by pressing the 
physical AC button on the Delta 2 unit. After that the switch is 
available and fully controllable via Home Assistant.

### Integration version

0.9.1

### Home Assistant version

2026.5.4

### Device model

Delta 2  Firmware V1.0.2.176

### Bluetooth connection

Internal bluetooth adapter

No

```text
Home Assistant is running in a Docker environment on a Raspberry Pi 5 (16GB) with Debian Linux: 
Linux Raspi516 6.18.29+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.18.29-1+rpt1 (2026-05-12) aarch64 GNU/Linux
```

### Diagnostics

_No response_

### Relevant log output

```shell

```

GnoX:
The inverter heartbeat is the only message that carries `cfg_ac_enabled` (which drives the `ac_ports` field), and the firmware stops sending it entirely while AC output is off. After a fresh connect with AC off, `ac_ports` is therefore never populated, the switch reports unavailable (its value stays `None`), and HA blocks `turn_on` on unavailable entities - so the AC ports can't be enabled from HA at all until the inverter is turned on physically or via the app.

`Delta2Base` now arms a 10 second fallback when the device authenticates: if no inverter heartbeat arrives within the window, `ac_ports` falls back to off so the switch becomes controllable. It is deliberately left unavailable during the grace period instead of defaulting immediately, so we never assume off (and send an `on` command) for a port that is actually on but slow to report. A real heartbeat at any point still overrides the fallback.

Resolves #351, resolves #334

### DIFF
--- a/custom_components/ef_ble/eflib/devices/_delta2_base.py
+++ b/custom_components/ef_ble/eflib/devices/_delta2_base.py
@@ -1,3 +1,7 @@
+from bleak.backends.device import BLEDevice
+from bleak.backends.scanner import AdvertisementData
+
+from ..connection import ConnectionState
 from ..devicebase import DeviceBase
 from ..entity import controls
 from ..entity.base import dynamic
@@ -85,6 +89,26 @@ class Delta2Base(DeviceBase, RawDataProps):
     dc_12v_port = raw_field(pb_pd.car_state, lambda x: x == 1)
     dc12v_output_voltage = raw_field(pb_mppt.car_out_vol, pdiv(1000, 2))
     dc12v_output_current = raw_field(pb_mppt.car_out_amp, pdiv(1000, 2))
+
+    def __init__(
+        self, ble_dev: BLEDevice, adv_data: AdvertisementData, sn: str
+    ) -> None:
+        super().__init__(ble_dev, adv_data, sn)
+        self.on_connection_state_change(self._default_ac_ports_off_when_missing)
+
+    def _default_ac_ports_off_when_missing(self, state: ConnectionState) -> None:
+        # The inverter stops sending its heartbeat entirely while AC output is off, so
+        # `ac_ports` never gets a value after a fresh connect and the switch stays
+        # `unavailable`. Default it to off if nothing arrives within 10s - a real
+        # heartbeat afterwards still overrides this.
+        if state != ConnectionState.AUTHENTICATED:
+            return
+
+        def _set_off_if_unknown() -> None:
+            if self.ac_ports is None:
+                self.notify_field(Delta2Base.ac_ports, False)
+
+        self.call_later(10, _set_off_if_unknown, key="default_ac_ports_off")
 
     @property
     def pd_heart_type(self):


rabits (approved):
Great fix!

PR #357, opened 2026-05-29
Ukrainian translation

megadimich:
Ukrainian translation


custom_components/ef_ble/translations/uk.json
GnoX, line 1: Country code 'uk' is not valid, this should be 'ua.json'.

PR #359, opened 2026-05-29
Fix config entry FAILED_UNLOAD on hung BLE disconnect, add disconnect tracing

### LINKED ISSUE #338: [Bug]: SHP2/DPU enters ConfigEntryState.FAILED_UNLOAD during BLE flap — async_unload_entry blocks 20 s on BleakClient.disconnect()

GnoX:
`BleakClient.disconnect()` can block until the connect timeout (default 20s) when a write-with-response is still pending on the transport after a mid-auth BLE drop, which is common through an ESPHome proxy. `Connection.disconnect()` awaited it with no deadline, so `async_unload_entry` could outrun HA's unload window and leave the entry in `ConfigEntryState.FAILED_UNLOAD` - which only a full restart recovers from, re-triggering the same flap.

Every client teardown now goes through a bounded `_disconnect_client()` helper (`asyncio.timeout`, `DISCONNECT_TIMEOUT = 5s`) that swallows the usual "already down" errors and, on timeout, gives up waiting and lets the caller finish local cleanup - the transport drains on its own once the drop is detected. This enforces the deadline in the library where the hang actually happens, so all five teardown paths benefit and no `async_unload_entry` guard is needed.

While here, make disconnects traceable from a diagnostics download alone (no debug logging required), since the original report was hard to pin down:
- `disconnect()` and the bleak `disconnected()` callback record the caller chain as the `reason` on the `DISCONNECTING` / `DISCONNECTED` entries in `connection_state_history`, so a requested disconnect (unload, reload) is distinguishable from an unsolicited bleak drop.
- `_disconnect_client()` records each outcome (`ok` / `timeout` / `already_down`) with its trigger into a bounded `disconnect_log` that is included in the diagnostics dump, so timeouts and which path caused them are visible across every teardown path.

Resolves #338. Reimplements #339 with the fix moved into the library and the root cause framed correctly.
Should also address #366 #367

### DIFF
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -192,7 +192,10 @@ async def async_unload_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> b
     """Unload a config entry."""
     _cancel_reappear_callback(hass, entry)
     device = entry.runtime_data
-    await device.disconnect()
+    try:
+        await device.disconnect()
+    except Exception:
+        _LOGGER.exception("Error disconnecting device during unload, continuing")
     device.with_logging_options(LogOptions.no_options())
     return await hass.config_entries.async_unload_platforms(entry, PLATFORMS)
 
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -507,6 +507,14 @@ class Connection:
                 DISCONNECT_TIMEOUT,
                 trigger,
             )
+        except (OSError, RuntimeError) as e:
+            outcome = f"transport_broken: {e}"
+            self._logger.warning(
+                "BleakClient.disconnect() raised %s (%s); the BLE transport is broken, "
+                "continuing with local cleanup",
+                type(e).__name__,
+                trigger,
+            )
         self._disconnect_log.append(
             {"time": time.time(), "trigger": trigger, "outcome": outcome}
         )
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -11,7 +11,7 @@ from collections.abc import Awaitable, Callable, Collection, Coroutine, MutableS
 from dataclasses import dataclass
 from enum import StrEnum, auto
 from functools import cached_property
-from typing import Literal, Self
+from typing import Any, Literal, Self
 
 import ecdsa
 from bleak import BleakClient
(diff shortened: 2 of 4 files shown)


rabits (approved):
Thank you! Looks great

PR #360, opened 2026-05-29
Fix spurious SHP2 error logging, clean up packet parsing

### LINKED ISSUE #340: [Bug]: SHP2 logs spurious 'Error happened on device: []' on every BLE reconnect
## Summary

`shp2.py` fires a `WARNING` log saying **"Error happened on device: []"** on **every BLE reconnect**, even when there are no actual device errors. This is a false positive that pollutes logs and (depending on HA notification automations) can trigger false alerts.

## Root Cause

In `data_parse()` (around line 290):

```python
self.error_count = len(self.errors) if self.errors is not None else None

if (
    self.error_count is not None
    and prev_error_count is not None
    and self.error_count > prev_error_count
) or (self.error_count is not None and prev_error_count is None):
    self.error_happened = True
    self._logger.warning("Error happened on device: %s", self.errors)
```

The second branch:
```python
(self.error_count is not None and prev_error_count is None)
```
fires whenever **`error_count` transitions from `None` → any value** — including `0`.

On every BLE reconnect, the first `backup_incre_info` packet received causes `self.errors` to be set from `None` to `[]` (empty list), making `error_count = 0`. Since `prev_error_count` was `None` (no previous data for this session), the condition is satisfied and the warning fires — even though the errors list is **empty**.

## Observed Behavior

Every BLE reconnect cycle (the SHP2 firmware drops the BLE connection roughly every 10 s) produces:
```
WARNING custom_components.ef_ble.eflib.devices.shp2 - 2C:BC:BB:XX:XX:XX  Error happened on device: []
```

With a −59 dBm proxy at 2 m distance cycling continuously, this produces ~6 false warnings per minute.

## Proposed Fix

Add a guard for `error_count > 0` — an empty error list is not an error condition:

```python
if (
    self.error_count is not None
    and self.error_count > 0
    and (prev_error_count is None or self.error_count > prev_error_count)
):
    self.error_happened = True
    self._logger.warning("Error happened on device: %s", self.errors)
```

This retains correct detection of real errors (non-empty `errcode` list that grows) while silencing the spurious `None → 0` transition.

## Environment

- ef_ble v0.9.0 (HACS)
- Device: Smart Home Panel 2 (SHP2) — also likely affects any device that uses the same pattern
- Proxy: ESPHome BLE proxy (ESP32-POE, wired Ethernet)
- HA 2025.5.x on NixOS

## Related

- Issue #338 (BLE flap / FAILED_UNLOAD) — the reconnect cadence that triggers this warning is a separate known characteristic of SHP2 firmware.

GnoX:
SHP2 logged a `WARNING` "Error happened on device: []" even when the device reported no errors. `error_count` transitions `None -> 0` the first time an errcode list is parsed - an all-zero errcode filters to `[]`, so the count is `0` - and the `prev_error_count is None` branch fired regardless of whether the list was empty, also flipping the `error_happened` problem sensor on.

Now we warn only when the error count actually grows beyond the snapshot taken before the packet updated `errors`; an empty errcode list is never treated as an error.

`error_count` and `error_happened` are now computed fields derived from `errors`, so they share a single source of truth and can't drift out of sync. This also fixes a latent issue where `error_happened` was set `True` and never reset - leaving the problem binary sensor stuck "on" forever after any error - it now reflects the current state (on iff the device currently reports errors).

While here, `data_parse` packet routing is aligned with the other devices by replacing the nested `if/elif` chain with a `match (src, cmd_set, cmd_id)` block.

Resolves #340.

### DIFF
--- a/custom_components/ef_ble/eflib/devices/shp2.py
+++ b/custom_components/ef_ble/eflib/devices/shp2.py
@@ -8,8 +8,8 @@ from ..entity.base import dynamic
 from ..packet import Packet
 from ..pb import pd303_pb2
 from ..props import (
-    Field,
     ProtobufProps,
+    computed_field,
     field_group,
     pb_field,
     pb_group,
@@ -203,8 +203,14 @@ class Device(DeviceBase, ProtobufProps):
     ac_charging_speed = pb_field(pb_push_set.charge_watt_power)
 
     errors = pb_field(pb_push_set.backup_incre_info.errcode, _errors)
-    error_count = Field[int]()
-    error_happened = Field[bool]()
+
+    @computed_field
+    def error_count(self) -> int | None:
+        return len(self.errors) if self.errors is not None else None
+
+    @computed_field
+    def error_happened(self) -> bool:
+        return bool(self.errors)
 
     @staticmethod
     def check(sn: bytes):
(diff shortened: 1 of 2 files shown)


rabits (approved):
Looks good!

PR #361, opened 2026-05-30
Fix AC sensors stuck unavailable while AC output is off for River 2 and Delta 2

GnoX:
This is a followup of #356 - in there, we only updated switch, but all other fields are also unavailable so this PR adds a functionality that allows setting defaults for all fields. This also addresses a comment in https://github.com/rabits/ha-ef-ble/issues/351#issuecomment-4583005082 that reported this issue happens for River 2.

This adds a small, generic mechanism rather than patching one field. A field can declare its off value with `Field.default_when_missing(value)`. When a device has any such field, `DeviceBase` schedules a one-shot fallback after authentication (`MISSING_DEFAULT_GRACE = 10s`): for each marked field that is still unset once the grace period elapses, it applies the declared value through `notify_field`. A real message arriving at any point - before or after - still wins. The field iteration and a `TypeIs` guard live on `UpdatableProps` (`is_props` / `fields_with_missing_default`), so `DeviceBase` never reaches into props internals.

Config / spec fields (configured charge speed, rated power, etc.) are deliberately left unmarked - an empty inverter reading should not clobber a setpoint to `0`, so those stay unavailable until the device actually reports them.

### DIFF
--- a/custom_components/ef_ble/eflib/devicebase.py
+++ b/custom_components/ef_ble/eflib/devicebase.py
@@ -29,7 +29,12 @@ from .logging_util import (
 )
 from .packet import Packet
 from .props.raw_data_props import Literal
-from .props.updatable_props import Field
+from .props.updatable_props import Field, UpdatableProps
+
+# Seconds to wait after authentication before falling back to a field's
+# `default_when_missing` value (covers devices that withhold a whole message while the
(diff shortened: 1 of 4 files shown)


rabits (approved):
Yeah looks good!

Issue #375, opened 2026-06-12
[Bug]: Keeps disconnecting from Smart Home Panel 2


GHJeffE30:
### Pre-submission checks

- [x] I'm running the latest release and the bug still reproduces
- [x] I searched existing issues and this isn't a duplicate

### What happened?

My BLE just stops updating several items.  It keeps trying to reinitialize, but most of the time, can't.  It's missing items like grid status, grid connection, power output, etc.  I have the latest release you put out, but Home Assistant had updated twice, and the Smart Home Panel 2 has updated to version 4.03.153 in the last few days.  The problem was happening before then.  The only way to get it to stop is to restart Home Assistant.  That solves it for a few hours, then it starts again.  Let me just say thank you for your work on making this.  It's been a great tool for me to use!  Thanks you

### Integration version

0.9.3

### Home Assistant version

2023.6.2

### Device model

SHP2 v 4.03.153

### Bluetooth connection

Internal bluetooth adapter

### Have you changed the integration settings?

No

### Changed integration settings

```text

```

### Diagnostics

When i tried to get it to reconnect, it gave one error of unknown.  Part is not in list, or something like that.  Another error it gave was didn't make it to auth, timeout

### Relevant log output

```shell

```

GHJeffE30:
I should also tell you that the Delta Pro Ultra works just fine.  This is only on the Smart Home Panel 2

rabits:
Hi @GHJeffE30, the SHP2 are quite demanding in regard for bandwidth. I do not update my setup, but the first thing to check is the logs, and if you see some protocol issues (like crc ones) - I would recommend to change the BLE device, for example ESP32 ETH works just fine for my 2xSHP2 panels (https://github.com/rabits/ha-ef-ble/wiki#i-frequently-see-that-device-is-disconnected-or-log-shows-alot-of-exceptions-about-crcpadding). If there is something else - will be great if you can attach the logs.

GHJeffE30:
Here is what I got this morning after a reboot of home assistant.


This error originated from a custom integration.

Logger: custom_components.ef_ble.eflib.connection - EC:C9:FF:CD:22:7E
Source: custom_components/ef_ble/eflib/connection.py:636
Integration: EcoFlow BLE (Unofficial) (documentation, issues)
First occurred: 7:58:30 AM (6 occurrences)
Last logged: 7:58:30 AM

[...]

Issue #367, opened 2026-06-06
[Bug]: Initializing for 5 hours


jan-tdy:
### Pre-submission checks

- [x] I'm running the latest release and the bug still reproduces
- [x] I searched existing issues and this isn't a duplicate

### What happened?

All entities stopped being avalaible 5 hourss ago. The integration is initializing.

### Integration version

0.92

### Home Assistant version

2026.06.0

### Device model

River 3 Plus

### Bluetooth connection

Internal bluetooth adapter

### Have you changed the integration settings?

No

### Changed integration settings

```text

```

### Diagnostics

Not possible, still initializing.

### Relevant log output

```shell
This error originated from a custom integration.

Logger: custom_components.ef_ble.eflib.connection - E4:B0:63:56:BC:C6
Source: custom_components/ef_ble/eflib/connection.py:595
Integration: EcoFlow BLE (Unofficial) (documentation, issues)
First occurred: 4:53:05 AM (1 occurrence)
Last logged: 4:53:05 AM

Captured exception: [org.freedesktop.DBus.Error.NoReply] Remote peer disconnected: File "/config/custom_components/ef_ble/eflib/connection.py", line 764, in sendRequest await self._sendRequest(send_data, response_handler) File "/config/custom_components/ef_ble/eflib/connection.py", line 803, in _sendRequest await self._start_notify(response_handler) File "/config/custom_components/ef_ble/eflib/connection.py", line 790, in _start_notify await self._client.start_notify(self._notify_characteristic, callback, **kwargs) File "/usr/local/lib/python3.14/site-packages/bleak/__init__.py", line 904, in start_notify await self._backend.start_notify( characteristic, wrapped_callback, bluez=bluez, cb=cb, **kwargs ) File "/usr/local/lib/python3.14/site-packages/bleak/backends/bluezdbus/client.py", line 988, in start_notify assert_gatt_reply(reply, start_notify=True) ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.14/site-packages/bleak/backends/bluezdbus/utils.py", line 112, in assert_gatt_reply assert_reply(reply) ~~~~~~~~~~~~^^^^^^^ File "/usr/local/lib/python3.14/site-packages/bleak/backends/bluezdbus/utils.py", line 34, in assert_reply raise BleakDBusError(reply.error_name, reply.body)





This error originated from a custom integration.

Logger: homeassistant
[...]

PR #377, opened 2026-06-14
Fix crash on BLE packets received after disconnect, throttle repeated time-sync requests

GnoX:
This PR fixes two independent BLE connection-lifecycle issues that surfaced as repeated reconnect flapping and noisy error logs on long-running setups.

- **`'NoneType' object has no attribute 'replyPacket'` after disconnect.** `listenForDataHandler` is registered directly as the bleak notify callback, so it isn't tracked in `_tasks` and `_cancel_tasks()` never cancels it on teardown. bleak can still hand us notifications buffered before the drop *after* `Connection.disconnect()` has cleared `_client` and the device has cleared its `_conn`. The handler now bails out of the packet loop as soon as `_client` is `None` - guarding the central choke point covers every device and every reply path, and replying on a dead link is pointless anyway. This could've been the cause for #375.

- **Repeated time-sync requests flooded the link.** Some devices (e.g. River 3 Plus in #367) re-request the time in a tight loop when they don't accept the response, which had us fire a fresh `sendUtcTime` / `sendRTCRespond` / `sendRTCCheck` burst on every request.  The device only needs the time set occasionally, so `TimeCommands.async_send_all` now collapses repeated requests within a 30s window into a single send

### DIFF
--- a/custom_components/ef_ble/eflib/commands.py
+++ b/custom_components/ef_ble/eflib/commands.py
@@ -1,7 +1,7 @@
 import logging
 import struct
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from .devicebase import DeviceBase
 from .packet import Packet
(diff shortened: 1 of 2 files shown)


rabits (approved):
Looks very good!

Issue #362, opened 2026-05-31
[Sensor]: Delta Pro Extra Battery reports 0% / 0°C when plugged into a Delta Pro 3


xthexder:
### Pre-submission checks

- [x] I'm running the latest release and the issue still reproduces
- [x] I searched existing issues and this isn't a duplicate

### What kind of change are you asking for?

Wrong value (sensor exists but is incorrect)

### Device model

Delta Pro Extra Battery

### Sensor / control name

Battery Level / Cell Temperature

### What's wrong or what's needed?

Support for reading extra battery levels was added recently, and I reconfigured my setup to add my devices.
My Delta Pro 3 Extra Battery works fine and reports correctly in home assistant, however my older Delta Pro Extra Battery connected to the same Delta Pro 3 reports as 0% and 0°C even though it is fully charged and around 25°C.

### Integration version

0.9.1

### Diagnostics

Diagnostics attached. 2 Extra Batteries are attached to this Delta Pro 3, the first of which is the older 3600Wh battery.

Rough battery levels during collection:
Built-in battery (4096Wh): 78% 27°C
Extra Battery 1 (3600Wh): 87% (unsure, roughly 25°C)
Extra Battery 2 (4096Wh): 95% 22°C

[config_entry-ef_ble-01KCWWABE4JHKDSFPHEQSEA7CR-1.json](https://github.com/user-attachments/files/28444791/config_entry-ef_ble-01KCWWABE4JHKDSFPHEQSEA7CR-1.json)

### Anything else?

Delta Pro Extra Batteries are compatible with both Delta Pro and Delta Pro 3, they just have a lower 3600Wh capacity. Ideally this capacity value could also be read so it can properly calculate the total remaining Watt-hours, and not just a percentage.

GnoX:
The second battery is in fact reporting that it is not active in the diagnostics you provided and there is nothing to indicate that it uses different field for this. CMS SOC (all batteries pooled together) is even reporting `86.88%` which is exactly average of main battery and extra battery (`(78.26 + 95.5) / 2 = 86.88`). If you can't see it in the app, we can't see it either.

xthexder:
I guess I should have disabled HA and checked in the Ecoflow app before submitting this, my batteries were not connected quite how I thought, and there was only the 3600Wh one on port 2.
Port 1 was disconnected even though it showed up during reconfiguration. (In my defense it's slightly harder to tell what the DP3 Battery is doing due to no display)

[...]

PR #378, opened 2026-06-14
Refresh extra battery serial number and model in the device registry

GnoX:
The serial number of an extra battery rides in only about half of the device's `DisplayPropertyUpload` messages, so it can arrive *after* the extra-battery entity - and its device registry entry - were first created. `EcoflowBatteryAddonEntity.device_info` exposes `serial_number` (and the model name derived from it via `battery_name_from_device`), but HA reads `device_info` only once when the entity is added. If no SN-bearing message had landed before platform setup, the registry entry was written with `serial_number=None` and a fallback `"Extra Battery"` model and then never refreshed - so the serial number was missing "in some cases" depending on timing.

This PR makes `EcoflowBatteryAddonEntity` register a state-update callback on `battery_{index}_sn` and pushes the value into the device registry (`async_update_device`) once it is known, updating both the serial number and the model. It is a no-op when the SN is empty, the registry entry is missing, or the entry is already current.

Should address discussion in #362

### DIFF
--- a/custom_components/ef_ble/entity.py
+++ b/custom_components/ef_ble/entity.py
@@ -3,6 +3,7 @@ from collections.abc import Callable, Mapping
 from typing import Any, Protocol, runtime_checkable
 
 from homeassistant.core import callback
+from homeassistant.helpers import device_registry as dr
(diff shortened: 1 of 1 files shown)


rabits (approved):
Looks good, thank you!

PR #385, opened 2026-06-22
Restore PowerStream battery level field, expose the other SoC fields as diagnostics

### LINKED ISSUE #369: [Sensor]: PowerStream Battery Level not reported anymore
### Pre-submission checks

### Device model

PowerStream, FW 1.0.1.288

### Sensor / control name

sensor.powerstream_battery_level

The battery level of the extra battery connected to the PowerStream is reported "Unknown" instead of the actual value. All other sensor values are reported via EcoFlow BLE. Via EcoFlow-Cloud the battery level value is reported. 

This issues started occurring 6 June 2026 @ 7:31:44, around that time I did the EcoFlow BLE updates* as well: 0.9.2 first and some time later 0.9.3, don't know exact timings.

I can't say for sure it is related to the updates, but it is clear the value of the battery level is currently not reported via EcoFlow BLE while it is via EcoFlow-Cloud. 

### Integration version

0.9.3

### Diagnostics

_No response_

### Anything else?

Image

GnoX:
#337 switched the PowerStream `battery_level` to `new_psdr_heartbeat.f32_lcd_show_soc` because it tracked the cloud / Delta 2 Max SoC most closely. That field turns out not to be reported by every PowerStream (e.g. the 600W on FW 1.0.1.288), so for those devices the battery level has been showing `Unknown` ever since - even though every other value is reported fine over BLE and the SoC is available via the cloud.

This restores the default `battery_level` to `f32_show_soc` (the field used before 0.9.1), which is present on all PowerStreams, and exposes the other SoC readings as diagnostic, disabled-by-default sensors so the value some users preferred isn't lost and the alternatives stay available for debugging:
- `lcd_battery_level` - `new_psdr_heartbeat.f32_lcd_show_soc` (what 0.9.1 used)
- `bms_battery_level` - `bat_soc`

Resolves #369

### DIFF
--- a/custom_components/ef_ble/eflib/devices/powerstream.py
+++ b/custom_components/ef_ble/eflib/devices/powerstream.py
@@ -34,7 +34,9 @@ class Device(DeviceBase, ProtobufProps):
     pv_current_2 = pb_field(pb.pv2_input_cur, pdiv(10, 1))
     pv_temperature_2 = pb_field(pb.pv2_temp, pdiv(10, 1))
 
-    battery_level = pb_field(pb_inv2.new_psdr_heartbeat.f32_lcd_show_soc, pround(2))
+    battery_level = pb_field(pb_inv2.new_psdr_heartbeat.f32_show_soc, pround(2))
+    lcd_battery_level = pb_field(pb_inv2.new_psdr_heartbeat.f32_lcd_show_soc, pround(2))
+    bms_battery_level = pb_field(pb.bat_soc)
     battery_power = pb_field(pb.bat_input_watts, pdiv(10, 1))
     battery_temperature = pb_field(pb.bat_temp, pdiv(10, 1))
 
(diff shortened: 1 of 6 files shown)


GnoX:
@rabits Aaand it happened, AI silently rewrote my remote and I accidentally pushed to your main, sorry. I knew it could happen so I was being really careful before every push, but I got lax after this weekend out. Would you be able to drop the last 2 commits? Not a big deal if you won't, it didn't change as I reverted the first commit with the second. Can you also able to enable branch protection on main against non-PR pushes?

rabits (approved):
Looks nice!

GnoX:
Looks like it's too late now to drop those commits, this was ready from the start, I was just waiting for reply. My other PR is ready too and didn't want to create SHP3 and PowerOcean PR's as I'm trying to limit it to two parallel ones.

rabits:
> @rabits Aaand it happened, AI silently rewrote my remote and I accidentally pushed to your main, sorry. I knew it could happen so I was being really careful before every push, but I got lax after this weekend out. Would you be able to drop the last 2 commits? Not a big deal if you won't, it didn't change as I reverted the first commit with the second. Can you also able to enable branch protection on main against non-PR pushes?

Oh missed that comment... Will check what's up there...

rabits:
@GnoX ok, added restrictions on the main branch.

PR #402, opened 2026-07-06
Reset SHP3 circuit power sensors to zero when a circuit stops drawing

### LINKED ISSUE #401: [Bug]: SHP3: circuit power sensors keep a stale value when a circuit stops drawing (should reset to 0)
### Pre-submission checks

### What happened?

Have an SHP3 circuit under load.
Turn that load off .
Watch that circuit's ..._circuit_power_* sensor.

Actual: power sensor continues to report last value
Expected: power sensor should report zero

### Integration version

1.0.0

### Home Assistant version

2026.7.1

### Device model

SHP3

### Bluetooth connection

ESPHome bluetooth proxy

No

```text

```

### Diagnostics

_No response_

### Relevant log output

```shell

```

GnoX:
On the Smart Home Panel 3 a `..._circuit_power_*` sensor kept reporting its last value after the circuit's load was switched off, instead of dropping to 0. The panel keeps sending each circuit's `load_ch{n}_sample_info` every heartbeat, but once a circuit stops drawing it **omits the `load_ch_power` field** while still sending `load_ch_vol` and `load_ch_curr`. Our `circuit_power` field extracts the nested `load_ch_power`, and with it absent the update was skipped, so the sensor held the stale non-zero value.

This PR fixes it by treating a missing `load_ch_power` as `0` via `TransformIfMissing` (the same mechanism `circuit_status` already uses for a missing `load_sta`), so the sensor resets to 0 instead of keeping the last reading

Resolves #401

### DIFF
--- a/custom_components/ef_ble/eflib/devices/shp3.py
+++ b/custom_components/ef_ble/eflib/devices/shp3.py
@@ -149,7 +149,9 @@ class Device(DeviceBase, ProtobufProps):
         pb.load_ch1_sample_info.load_ch_power,
         match="load_ch{n}_sample_info",
         count=NUM_OF_CIRCUITS,
-        transform=pround(2),
+        transform=TransformIfMissing[float, float](
+            lambda v: round(v, 2) if v is not None else 0.0
+        ),
         name_template="circuit_power_{n}",
     )
     circuit_current = pb_field_group(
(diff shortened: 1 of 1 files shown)


rabits (approved):
Yep looks nice!

PR #407, opened 2026-07-10
Fix Delta 3 Max Plus AC charging speed capped at 1500 W instead of 2400 W

### LINKED ISSUE #406: [Bug]: Delta 3 Max Plus AC charging speed capped at 1500W instead of 2400W
### Pre-submission checks

### What happened?

Steps to reproduce:
1. Connect a Delta 3 Max Plus (serial starts with D3M1) via the ef_ble integration
2. Open the "AC Charging Speed" number entity in Home Assistant
3. Try to set a value above 1500W (e.g. 1800W or 2400W)

Expected:
- Should allow up to 2400W, matching the official EcoFlow app

Actual:
- Home Assistant caps/rejects values above 1500W
- The EcoFlow app on the same device allows setting 2400W without issue

Device is correctly detected as "Delta 3 Max Plus" in the device info screen.

### Integration version

1.0.1

### Home Assistant version

2026.7.1

### Device model

Delta 3 Max Plus, Firmware 7.80.99.53

### Bluetooth connection

Internal bluetooth adapter

No

```text

```

### Diagnostics

[config_entry-ef_ble-01KS9DHYWDB1MGWWFZ9CM6KZVP.json](https://github.com/user-attachments/files/29908961/config_entry-ef_ble-01KS9DHYWDB1MGWWFZ9CM6KZVP.json)

Possible cause (for maintainers):
Delta3Base hardcodes max_ac_charging_power to 1500W via @computed_field.
delta3_max_plus.py tries to override with pb_field(plug_in_info_ac_in_chg_hal_pow_max),
but both use the same internal storage, so _recompute() overwrites the device-reported
limit back to 1500W on each BLE update.
Device details:
- Model: Delta 3 Max Plus
- SN prefix: D3M1
- Connection state: authenticated

### Relevant log output

```shell

```

GnoX:
On the Delta 3 Max Plus the "AC Charging Speed" number entity rejected anything above 1500 W even though the device charges at up to 2400 W. The subclass overrides the Delta 3 base's computed `max_ac_charging_power` (hardcoded `1500`) with a `pb_field` reading the device-reported `plug_in_info_ac_in_chg_hal_pow_max`, and while `Field.__set_name__` replaces the descriptor in `_fields`, the inherited computed descriptor **stayed registered in `_computed_fields`**. Both descriptors resolve to the same private storage slot, so `_recompute()` ran the stale base function after every heartbeat and overwrote the device-reported 2400 back to 1500.

This PR makes it possible to override computed fields by evicting them, but also ports usage of this behavior to computed fields themselves so this is more of a fix for potential future bugs. Tests that would catch this were also added for this device.

Resolves #406

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta3_max_plus.py
+++ b/custom_components/ef_ble/eflib/devices/delta3_max_plus.py
@@ -2,7 +2,7 @@ from collections.abc import Sequence
 
 from ..entity import controls
 from ..pb import pd335_sys_pb2
-from ..props import pb_field, repeated_pb_field_type
+from ..props import computed_field, pb_field, repeated_pb_field_type
 from ..props.transforms import flow_is_on, out_power
 from . import delta3_plus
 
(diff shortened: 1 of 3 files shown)


rabits (approved):
Yep looks good!

PR #408, opened 2026-07-11
Add power off button for River 3 and Delta 3 series

### LINKED ISSUE #384: [Enhancement Request]: Add device shutdown support for EcoFlow Delta 3 Plus
### Pre-submission checks

### Device model

Delta 3+

### Sensor / control name

Power Off

### Description

The EcoFlow Delta 3 Plus supports a **device shutdown command** in the official EcoFlow app. This feature is currently not exposed in the BLE Home Assistant integration.

(The Delta 2 does **not** have this capability) 

### Expected behavior

A switch or button entity should be available in Home Assistant to trigger the device shutdown of the EcoFlow Delta 3 Plus — mirroring the functionality available in the official EcoFlow app.

### Current behavior

No shutdown entity is exposed for the Delta 3 Plus via the BLE integration.

### Device

- **Model:** EcoFlow Delta 3 Plus
- **Integration:** EcoFlow BLE (unofficial) V. 0.9.4

### Additional context

This feature exists in the official EcoFlow app for the Delta 3 Plus but is absent e.g. on the Delta 2. It would be especially useful for automation scenarios where the device should be powered down remotely or on a schedule via Home Assistant.

### Integration version

0.9.4

### Diagnostics

_No response_

### Anything else?

_No response_

GnoX:
This PR adds a button for Delta 3 and River 3 series that shuts off the device using `cfg_power_off` config option (disabled by default).

Buttons had no place in the controls system (they're stateless, and every existing control type hangs off a state field), so this adds one:

- New `@controls.button()` control type: the decorator registers an internal field named after the method, so the entity key derives from the method name and the method stays directly callable on the device.
- The button platform is ported to the same auto-discovery pattern as switches/numbers/selects (`ButtonBuilder` + `get_controls`), replacing the hardcoded `BUTTON_TYPES` list.
- `power_off` is added to the Delta 3 base (covers Delta 3, Plus, Air, Max, Ultra variants) and River 3 (covers Plus). The entity is **disabled by default** since it's a hard shutdown - users opt in from the entity registry.
- DPU's `unpause_solar` is migrated to the new control type, removing the deprecated button list entirely. The entity key is unchanged, so existing registrations keep working.

Resolves #384

### DIFF
--- a/custom_components/ef_ble/button.py
+++ b/custom_components/ef_ble/button.py
@@ -20,14 +20,8 @@ from .entity import EcoflowEntity
 
 @dataclass(frozen=True, kw_only=True)
 class EcoflowButtonEntityDescription(ButtonEntityDescription):
-    press_func: Callable[[DeviceBase], Awaitable] | None = None
+    press_func: Callable[[DeviceBase], Awaitable]
     availability_prop: str | None = None
-
-
-BUTTON_TYPES = [
-    # DPU
-    EcoflowButtonEntityDescription(key="unpause_solar"),
-]
 
 
 @dataclass(init=False)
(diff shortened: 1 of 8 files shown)


rabits (approved):
Looks great! Thank you

Issue #403, opened 2026-07-06
[Bug]: River 3 Plus — "Failed to set up" on firmware V1.33.81.61, no error in logs even at debug


everykindofpeople:
### Pre-submission checks

- [x] I'm running the latest release and the bug still reproduces
- [x] I searched existing issues and this isn't a duplicate

### What happened?

Device: EcoFlow River 3 Plus
Firmware: V1.33.81.61 (updated same-day, out of the box, before ever attempting BLE pairing)
Integration version: v1.0.0 (via HACS)
Home Assistant: Core 2026.7.1, Home Assistant OS 18.1, Raspberry Pi 4 (rpi4-64)
Bluetooth: Onboard Pi 4 adapter, device ~3 feet away, no case/enclosure interference
Symptom:
The integration correctly discovers the device (EF-R32094) and the account login flow succeeds — User ID resolves automatically from email/password with no errors. However, the config entry immediately shows "Failed to set up" with no further detail in the UI.
Troubleshooting already done:
Deleted and re-added the integration 4+ times
Reloaded the failed entry multiple times
Force-quit the EcoFlow Android app and disabled its Bluetooth to rule out the single-BLE-connection-slot limitation
Power-cycled the Pi (ha host reboot) to rule out a Linux Bluetooth stack issue
Attempted to power-cycle the EcoFlow itself (holding the power button 3+ seconds did not appear to power it off, possibly due to active AC pass-through load)
Confirmed proximity/line-of-sight is not an issue (~3 ft, no obstructions)
Enabled debug logging (custom_components.ef_ble) — only logged Init EcoFlow BLE Integration on each attempt, with no further output, error, or traceback of any kind, even at debug level

Log15:18:34.893 DEBUG (Ms (full debug output for a setup attempt):
2026-07-06 ainThread) [custom_components.ef_ble] Init EcoFlow BLE Integration
(No further log lines appear between this and the entry showing "Failed to set up" in the UI.)

Question: Is firmware V1.33.81.61 known to be compatible with the current BLE protocol implementation? Given this is a same-day-updated unit that's never successfully connected, I suspect this may be a firmware/protocol mismatch rather than an environmental issue, but wanted to rule out the more common causes first (see above) before assuming that.
Happy to collect diagnostic data or run further tests if that's helpful — let me know what's most useful.

### Integration version

v1.0.0 

### Home Assistant version

[...]

PR #410, opened 2026-07-12
Fix silent "Failed to set up" on entries missing a user ID, show live connection progress in the config flow

GnoX:
In #403 a River 3 Plus discovers and logs in fine, but the entry immediately shows "Failed to set up" with nothing in the log even at debug. The root cause is `async_setup_entry` returning `False` when `entry.data` has no `user_id` - HA renders that as a permanent SETUP_ERROR with zero log output. Such an entry could actually be persisted: the user ID form field is optional, and `device_confirm`/`unsupported_device` created the entry directly after a successful cloud login without the validated ID ever making it into the entry data.

Full list of changes:
- **Entries can no longer be persisted without a user ID.** `_create_entry` now writes the validated `self._user_id` explicitly instead of relying on the optional form field round-tripping through `user_input`.
- **Missing data fails loudly.** The silent `return False` is replaced by `ConfigEntryError` with a translated message naming what is missing, so already-broken entries tell the user to remove and re-add the device instead of failing blank.
- **The connection test shows live progress instead of a blocking spinner.** Submitting the form hands off to a progress step that runs the test in the background and reports the real `ConnectionState` as it advances - establishing Bluetooth connection, exchanging encryption keys, authenticating - via a `ProgressPhase` enum whose members carry their connection states, with phase transitions observed through `observe_connection()`. On failure the original form returns with the mapped error; closing the dialog mid-test cancels the task and disconnects cleanly after it unwinds.
- **The progress steps are re-entrant by design.** The flow manager schedules a configure call per completed progress task and the frontend walks `show_progress_done` into the next step on its own, so `connect` and `connect_done` tolerate concurrent and stale invocations: the running phase task is reused unless the phase changed, the connection result outlives the task, and only the first successful invocation creates the entry (later ones abort as `already_configured`).


rabits (approved):
Great fix, thank you!

PR #411, opened 2026-07-12
Add energy monitoring mode select for STREAM devices

GnoX:
This PR adds switch for the STREAM system's energy delivery strategy between "Smart Meter Monitoring (zero feed-in)" and "Semi-Automated Monitoring" from Home Assistant, e.g. to allow grid export when the batteries are full without reaching for the app.

Full list of changes:
- New `EnergyMonitoringMode` enum and `energy_monitoring_mode` select on the STREAM AC base, inherited by all BK-series variants (AC / Pro / Max / Ultra / AC Pro).  The write reuses the existing `_send_config_packet` framing shared by the other STREAM config writes.
- Devices that were never configured report `0`; that maps to `UNKNOWN`, which is excluded from the select options and ignored as a set target.

Resolves #393

### DIFF
--- a/custom_components/ef_ble/eflib/devices/stream_ac.py
+++ b/custom_components/ef_ble/eflib/devices/stream_ac.py
@@ -126,6 +126,13 @@ class EnergyStrategy(IntFieldValue):
         return operate_mode
 
 
+class EnergyMonitoringMode(IntFieldValue):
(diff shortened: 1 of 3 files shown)


rabits (approved):
Looks good!

PR #415, opened 2026-07-12
BLE connection resilience improvements (ESP32 proxy focus)

ecammit:
This was an attempt to address some of the instability I was having when re-introducing esp32 ble_proxy connections to DPU and SHP2 (which seem to work now after the CRC performance improvements... or potentially other changes since then). PR #413 was a big part of fixing the instability, but this improves it further.

This was tested using two ESP-32 WROOM modules for ble_proxy, an Edimax bluetooth 5 USB adapter (rtl8761bu), and a new Krisnorey Bluetooth 6 USB Adapter (also rtl8761bu). It was tested connecting to two DPUs, an SHP2, and a Wave 2 while also connecting to several other bluetooth LE devices.

Screenshot 2026-07-12 at 22 12 18

Generally only the EcoFlow devices maintained persistent connections, however, all of them (SHP2, and DPU especially) are working on ESP-32 ble_proxy.

## What changed

Four incremental improvements to BLE connection resilience, each in its own commit with dedicated unit tests. Builds on top of #413 (already merged).

1. **Refresh the BLE device from HA's bluetooth manager on each connect attempt** - `Connection`/`DeviceBase` now accept an injectable `BLEDevice` resolver, wired in `async_setup_entry` to `bluetooth.async_ble_device_from_address`. Reconnects can now fail over to a healthier adapter/proxy instead of retrying a stale cached device.
2. **Distinguish "out of connection slots" from generic Bleak errors** - new `ConnectionState.ERROR_OUT_OF_SLOTS` for `bleak_retry_connector.BleakOutOfConnectionSlotsError`, plus a persistent repair issue during setup pointing at the real cause (ESP32 proxies typically support only ~3 concurrent slots).
3. **Capped exponential backoff with jitter for in-session reconnects, plus a configurable "Max reconnect attempts" option** - `MAX_RECONNECT_ATTEMPTS` default 2 -> 5, replaced flat `+10s` linear growth with capped exponential backoff (10s -> 20 -> 40 -> 60, capped) plus up to 20% jitter to avoid synchronized retries against a shared proxy. The attempt limit is now also exposed as an Advanced connection options preference.
4. **Watchdog for silently dead connections, gated by an Advanced connection option** - tracks time since the last received notification; if none arrives within 45s while otherwise "connected," forces a clean disconnect through the existing `_disconnect_client()` path so it's picked up by the normal reconnect logic. New "Enable connection watchdog" preference (defaults on).

## Why

ESP32 BLE proxies (ESPHome `bluetooth_proxy`) have failure modes distinct from a local BlueZ adapter - slot exhaustion, proxy flapping, and silent link loss - that the integration didn't previously handle well. These changes target those cases specifically without touching the core auth/connection handshake.

## Notes for reviewers

- Each commit is independently reviewable.
- New tests live in `tests/eflib/test_connection_resilience.py`, following the existing `tests/eflib` pattern of avoiding a real `homeassistant` import (mocking `bleak`/`bleak_retry_connector` directly instead).
- Two new Advanced connection options preferences (`max_reconnect_attempts`, `watchdog_enabled`); both are read live on each check/call, so toggling them doesn't require a reload.


custom_components/ef_ble/eflib/connection.py
```diff
@@ -48,15 +50,35 @@
 from .packet import Packet
 from .props.utils import classproperty
 
-MAX_RECONNECT_ATTEMPTS = 2
+MAX_RECONNECT_ATTEMPTS = 5
```
GnoX, line 53: This work on reconnect logic is a bit unnecessary as we always use `with_reconnect_disabled` so it's not even used for anything right now. It's probably ok to have it here for future, but it does nothing right now.

However, you added it as a config option - that we should remove because it has no effect.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -268,6 +296,8 @@ def __init__(
         self._state_reason: str | None = None
         self._set_state(ConnectionState.CREATED)
 
+        self.add_timer_task(self._watchdog_check, WATCHDOG_CHECK_INTERVAL)
```
GnoX: This changes the contract of `Connection` class - to construct it, it now needs to always run in an event loop otherwise it fails which I don't think we want. It's better to do this in connect.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -48,15 +50,35 @@
...
 # write-with-response is still pending on the transport after a mid-auth BLE drop
 # (notably through an ESPHome proxy). Left unbounded it stalls `async_unload_entry`
 # long enough for HA to mark the entry `FAILED_UNLOAD`, so cap every disconnect.
 DISCONNECT_TIMEOUT = 5.0
 
+# A proxy (or the OS Bluetooth stack) can report the link as connected while the
+# peripheral has actually gone silent without telling us. Force a reconnect if no
+# notification has arrived in a while rather than waiting on it indefinitely.
+WATCHDOG_CHECK_INTERVAL = 15.0
+WATCHDOG_TIMEOUT = 45.0
```
GnoX, line 80: While this sounds like a sensible value, I think it should be user-configurable. A device may appear that has some standby mode that only sends updates in higher intervals than this.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -253,6 +280,7 @@ def __init__(
 
         self._tasks: set[asyncio.Task] = set()
         self._call_later_handles: dict[str, asyncio.TimerHandle] = {}
+        self._last_activity: float = 0.0
```
GnoX, line 284: This could break some upcoming devices and make them instantly disconnect because they set AUTHENTICATED before they get to listenForDataHandler (where it's being updated right now). Watchdog would instantly kill them on the first tick. This should be updated when it sucessfully connects.

custom_components/ef_ble/__init__.py
```diff
@@ -198,6 +211,39 @@ async def _disconnect_and_reload():
...
+async def _handle_unauthenticated_connection(
+    hass: HomeAssistant,
+    entry: DeviceConfigEntry,
+    device: eflib.DeviceBase,
+    state: ConnectionState,
+    out_of_slots_issue_id: str,
+) -> None:
+    await device.disconnect()
+    if state is ConnectionState.ERROR_OUT_OF_SLOTS:
+        _create_out_of_slots_issue(hass, out_of_slots_issue_id, device.name)
```
GnoX, line 221: I did notice one thing that I forgot to add for the first issue - it should be deleted in `async_remove_entry` otherwise it will outlive the config entry.

GnoX (changes requested):

ecammit:
Re-testing modifications.

GnoX:
Can I ask why did you decide to close this? I think that at least the watchdog feature would be helpful for few consumers - there were few reports of silent link drops that required manual reloads.

ecammit:
> Can I ask why did you decide to close this? I think that at least the watchdog feature would be helpful for few consumers - there were few reports of silent link drops that required manual reloads.

@GnoX I wanted to start fresh on this given the current state has drifted significantly since this was written. As of now it's not at the top of my priority list either, so I didn't want a stale PR sitting open giving hope that I would get to it either.

PR #416, opened 2026-07-12
Fix Wave 2 Fahrenheit handling and drain controls, add climate entity

### LINKED ISSUE #327: [Bug]: Wave 2 Temp Faulty Celcius/Fahrenheit Conversion
### Pre-submission checks

### What happened?

I just installed the integration via HACS (v0.9.0) and am using it with a Wave 2. If the temperature settings on the Wave 2 are set to fahrenheit, the integration assumes the raw values are still in celsius and converts it. It seems to work correctly if I change the Wave 2 so it's in celsius mode (press and hold temp - and temp + buttons to toggle the f/c mode). As long as the Wave 2 is in celsius mode (and displays the temps on it's display in celsius) then the temps are accurate in home assistant.

To illustrate, the temp range for celsius is 16-30 degrees and for fahrenheit it's 60-86. If the Wave 2 is in fahrenheit mode (displaying temps in F on it's display) then in home assistant it'll show 140-187.0 F (it assumes 60F is 60C and converts 60C to 140F). It seems this breaks the integration being able to set the temp as it sends the wrong temp values to the Wave 2 in this configuration because even though the wave 2 is in fahrenheit mode, when changing the temperature in home assistant, the wave 2 will quickly show the weirdly converted celsius number for a second before clamping it to the lowest F temp (in F mode, change temp to lowest setting in HA and 16F flashes for like 300ms on the Wave 2 before changing back to 60F). It always clamps to 60F since all the weirdly converted numbers would be 16-30F when sending the temp back to the Wave 2.

### Integration version

v0.9.0

### Home Assistant version

2026.4

### Device model

Wave 2

### Bluetooth connection

ESPHome bluetooth proxy

No

```text

```

### Diagnostics

_No response_

### Relevant log output

```shell

```

GnoX:
This PR fixes three Wave 2 problems: target temperature was garbled when the unit is switched to Fahrenheit on the device, the drain mode/automatic drain controls neither reflected nor reliably controlled the actual drain state, and an internal "init" power mode leaked into the power mode select after the controls-system migration.

Full list of changes:
- **Temperature unit follows the device.** `temp_sys` is decoded into a `temp_unit` computed field, and the target temperature control gets `dynamic()` min/max/unit (16-30 °C or 60-86 °F). The number builder and climate entity learned to resolve dynamic units and limits per device, and the Wave 2 temperature sensors reuse the Wave 3 unit-switching helper.
- **Some Wave 2 frames carry a `0xBBBB` placeholder instead of a valid CRC16**; the packet parser now accepts that sentinel instead of dropping the frame.
- **Drain state is decoded mode-aware** (`automatic_drain`/`drain_mode` are computed from `wte_fth_en` + `main_mode`) and the writes mirror the app exactly: disabling always preserves the preference bit (`0b10 | bit0`), enabling in Heat/Fan always sends `1`, and selecting drain-free outside Cool mode is refused with a warning, matching the app's toast. A redundant drain write in `set_main_mode` (it always re-sent the value the device already had) was dropped.
- **`controls.select` gained an `exclude` parameter**, used to hide `PowerMode.INIT` from the power mode select the same way the pre-migration hardcoded select did. HA renders the state as unknown if the device ever reports it.
- **New climate entity for the Wave 2** (cool/heat/fan-only, low/medium/high fan, power mapped to `PowerMode.ON`/`STANDBY`), reusing the climate control introduced for Wave 3. Stacking a climate decorator over `@controls.select` on the same method previously lost the function name and broke dispatch; the control wrappers now use `functools.wraps`.
- Adds `tests/eflib/test_wave2.py` (first tests for this device) built on real captured heartbeats covering Heat/Fan modes, both temperature units and all four `wte_fth_en` values, plus exact command payloads for every drain write path.

Resolves #327

### DIFF
--- a/custom_components/ef_ble/climate.py
+++ b/custom_components/ef_ble/climate.py
@@ -15,7 +15,7 @@ from homeassistant.core import HomeAssistant, callback
 from homeassistant.helpers.entity_platform import AddConfigEntryEntitiesCallback
 
 from . import DeviceConfigEntry
-from .description_builder import EntityDescriptionBuilder
+from .description_builder import EntityDescriptionBuilder, unit_to_hassunit
 from .eflib import DeviceBase, controls, get_controls
 from .entity import EcoflowEntity
 
@@ -41,6 +41,8 @@ class EcoflowClimateEntityDescription(ClimateEntityDescription):
 
     min_temp: float | None = None
     max_temp: float | None = None
+    min_temp_prop: str | None = None
+    max_temp_prop: str | None = None
     target_temperature_step: float = 1.0
     min_range_temp: float | None = None
     max_range_temp: float | None = None
@@ -48,6 +50,7 @@ class EcoflowClimateEntityDescription(ClimateEntityDescription):
     min_humidity: int | None = None
     max_humidity: int | None = None
     temperature_unit: str = UnitOfTemperature.CELSIUS
+    temperature_unit_prop: str | None = None
 
     set_power_func: Any = None
     set_operating_mode_func: Any = None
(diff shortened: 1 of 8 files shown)


rabits (approved):
Great change!

PR #417, opened 2026-07-13
Expose battery voltage with min and max cell voltage for Delta 2 series

### LINKED ISSUE #409: [Feature Request] Internal battery voltage & cell voltages
Having migrated from the hassio-ecoflow-cloud integration to this, I was used to seeing the internal battery voltage as well as max & min cell voltages. These were rather useful to gauge if the reported battery percentage is less-than-accurate; if the minimum cell voltage was well above 3.0v yet the percentage was down near 10% or less, I knew the BMS calibration had wandered and could be readily drained down to 0% to re-calibrate it.

I'd love to see these entities over BLE on my Delta 2 Max (and any other device) given they're obviously seen inside the device itself, and are externally accessible - I just have no idea if there's an easy way to expose these values as-is or what work is required.

Even better would be to see every single cell voltage, however as I've not seen this anywhere I'm not entirely sure if this is accessible externally.

GnoX:
This PR adds internal battery pack voltage plus min/max cell voltage monitoring for the Delta 2 series, useful for spotting BMS calibration drift.

Full list of changes:

- **Fields live on `Delta2Base`**, so Delta 2, Delta 2 Max, Delta 2 Black and Delta 3 1500 all inherit them: `battery_voltage` (V, 2 decimals), `max_cell_voltage` / `min_cell_voltage` (V, 3 decimals), plus `battery_{1,2}_*` variants read from the per-battery BMS heartbeats for extra batteries.
- **`battery_voltage` reuses the existing "Battery Voltage" description** shared with DPU (diagnostic, disabled by default) and carries `max_cell_voltage` / `min_cell_voltage` as state attributes for a quick look without enabling more entities.
- **Min/max are also standalone diagnostic sensors** (disabled by default, 3-decimal precision), for history graphs - the frontend renders numeric attributes with at most 2 fraction digits, so attributes alone lose the millivolt resolution this feature exists for.
- **`_register_update_callback` gains an `entity_attr=None` mode** that just schedules a state write through the device's update-period throttle when the property changes. Sensors now declare their main field, state attribute fields and name field through it in `__init__`, and the base entity handles callback registration/removal - this also fixes state attributes staying frozen until the main sensor value happened to change.

Resolves #409


rabits (approved):
LGTM!)

PR #419, opened 2026-07-15
Add some Delta Pro 3 properties

MahdiBM:
**The changes in this PR are generated by Claude Fable (max effort). I can confirm these changes are effective on my device.**

Regarding the CONTRIBUTING.md, I can say most of the PR is simple so it's not like I don't understand most of the code, but I also can't claim I understand every single bit. I instructed Claude a bunch on different matters (e.g. just follow project conventions, make minimal changes, look at existing contributions, etc...) and I'm hoping it has done a good enough job. 

Below is what Claude has to say:

##

# Expand Delta Pro 3 support: battery diagnostics, port sensors, and switches

All new fields decode from the `DisplayPropertyUpload` messages already parsed for
the DP3 — no new packet types. Verified on real hardware (MR51, EU unit) by logging
~600 decoded uploads and checking values against the device.

**Sensors** (disabled by default, following #417 conventions):

| Entity | Proto field | Observed |
|---|---|---|
| State of Health | `cms_batt_soh` | 100 % |
| Battery Capacity | `cms_batt_full_energy` | 4096 Wh (nameplate) |
| Battery Charge State | `cms_chg_dsg_state` | see caveats |
| Min Cell Temperature | `bms_min_cell_temp` | 26 °C |
| Battery Input/Output Power | `pow_get_bms` | −18 W trickle → 18 W output; same split and translation keys as River 3 |
| Power In/Out Port Power | `pow_get_5p8` | 0 (port unused) |
| AC LV TT-30 Output Power | `pow_get_ac_lv_tt30_out` | 0 (port unused) |
| AC Output Frequency | `ac_out_freq` | 50 Hz |
| Fan Running (binary) | `pcs_fan_level` | same pattern as `_delta3_base` |

**Switches:** USB Ports (`cfg_usb_open`, mirrors `delta3.py`), Beeper
(`cfg_beep_en`), X-Boost (`cfg_xboost_en`). Beeper verified round-trip on hardware
(device echoed the change; no optimistic state involved). USB/X-Boost readback
matches the device; their writes use the same mechanism but weren't toggled.

**Also:** `icons.json` entries for the device-class-less entities (they rendered as
the eye icon), `en`/`uk`/`zh-Hans` translations, README table, and
`tests/eflib/test_delta_pro_3.py` — raw captured packets with exact-value
assertions. 153 tests pass, `prek run --all-files` green.

## Caveats

- **Battery Charge State** semantics from the official API docs (0 idle /
  1 discharging / 2 charging); our unit reported `0` throughout, even at 18 W
  trickle discharge — values 1/2 never observed. Disabled by default; can drop it
  if you prefer.
- **5P8 sign convention unverified** (nothing attached, constant 0) — exposed raw.
- **TT-30** always 0 on this EU unit; `out_power` transform by analogy with the
  other AC outputs.
- `uk`/`zh-Hans` translations not native-speaker reviewed.

## Deliberately not added

Never seen in any captured upload: `display_statistics_sum` (energy stats —
confirms the README note), `cms_batt_full/design/remain_cap`, `bypass_out_disable`,
`btn_usb_switch`, `cms_batt_temp` — entities would be permanently `unknown`.
Extra-battery port powers (`pow_get_4p8_*`) skipped: no packs here, and #250
suggests they read zero anyway. Volts/amps live in `RuntimePropertyUpload`
(cmd 0x16, unparsed) — possible follow-up.

### DIFF
--- a/custom_components/ef_ble/eflib/devices/delta_pro_3.py
+++ b/custom_components/ef_ble/eflib/devices/delta_pro_3.py
@@ -32,6 +32,14 @@ class DCPortState(IntFieldValue):
     STATE_5_UNKNOWN = 5
 
 
+class ChargeDischargeState(IntFieldValue):
+    UNKNOWN = -1
+
+    IDLE = 0
+    DISCHARGING = 1
+    CHARGING = 2
+
+
 class Device(DeviceBase, ProtobufProps):
     """Delta Pro 3"""
 
@@ -40,10 +48,17 @@ class Device(DeviceBase, ProtobufProps):
 
     battery_level = pb_field(pb.cms_batt_soc, pround(2))
     battery_level_main = pb_field(pb.bms_batt_soc, pround(2))
+    state_of_health = pb_field(pb.cms_batt_soh)
+    battery_full_energy = pb_field(pb.cms_batt_full_energy)
+    battery_charge_state = pb_field(
+        pb.cms_chg_dsg_state, ChargeDischargeState.from_value
+    )
 
     ac_input_power = pb_field(pb.pow_get_ac_in)
     ac_lv_output_power = pb_field(pb.pow_get_ac_lv_out, out_power)
     ac_hv_output_power = pb_field(pb.pow_get_ac_hv_out, out_power)
+    ac_lv_tt30_output_power = pb_field(pb.pow_get_ac_lv_tt30_out, out_power)
+    ac_output_frequency = pb_field(pb.ac_out_freq)
 
     input_power = pb_field(pb.pow_in_sum_w)
     output_power = pb_field(pb.pow_out_sum_w)
@@ -67,6 +82,9 @@ class Device(DeviceBase, ProtobufProps):
     plugged_in_ac = pb_field(pb.plug_in_info_ac_charger_flag)
     energy_backup = pb_field(pb.energy_backup_en)
     energy_backup_battery_level = pb_field(pb.energy_backup_start_soc)
+    battery_input_power = pb_field(pb.pow_get_bms, lambda value: max(0, value))
+    battery_output_power = pb_field(pb.pow_get_bms, lambda value: -min(0, value))
+    power_io_port_power = pb_field(pb.pow_get_5p8)
 
     battery_charge_limit_min = pb_field(pb.cms_min_dsg_soc)
     battery_charge_limit_max = pb_field(pb.cms_max_chg_soc)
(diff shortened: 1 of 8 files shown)


GnoX:
I can understand some sensors you added, but I'm intentionally avoiding some, can you tell me your use cases for:
- battery full energy - this is a static value, do you really need it as a sensor?
- battery charge state - can't you use battery power sensor for this?
- min cell temperature - what is the point of knowing both max and min temperatures?
- ac output frequency - isn't this also static value? either 50 or 60?
- xboost switch - is this something anyone even uses in automations? isn't this either permanently on or off?
- beeper switch - same as above

custom_components/ef_ble/eflib/devices/delta_pro_3.py
```diff
@@ -67,6 +82,9 @@ class Device(DeviceBase, ProtobufProps):
     plugged_in_ac = pb_field(pb.plug_in_info_ac_charger_flag)
     energy_backup = pb_field(pb.energy_backup_en)
     energy_backup_battery_level = pb_field(pb.energy_backup_start_soc)
+    battery_input_power = pb_field(pb.pow_get_bms, lambda value: max(0, value))
+    battery_output_power = pb_field(pb.pow_get_bms, lambda value: -min(0, value))
+    power_io_port_power = pb_field(pb.pow_get_5p8)
```
GnoX: This should be `ac_5p8_out_power` and `ac_5p8_in_power` as we already have this defined in sensors.

MahdiBM:
All valid points, I'm not using any of them.

For AC output freq, it seems like it's not static:
https://arc.net/l/quote/mroggerx

But I've removed it. I can re-add if needed.

custom_components/ef_ble/sensor.py
```diff
@@ -759,12 +759,16 @@ def shp2_circuit(
...
+    "ac_lv_tt30_output_power": power(precision=2, enabled=False),
     "solar_lv_power": power("input_power_solar_lv", enabled=False),
     "solar_hv_power": power("input_power_solar_hv", enabled=False),
     "dc_lv_input_power": power(precision=2),
     "dc_hv_input_power": power(precision=2),
     "dc_lv_input_state": enum(options=delta_pro_3.DCPortState),
     "dc_hv_input_state": enum(options=delta_pro_3.DCPortState),
+    "state_of_health": percentage(
+        enabled=False, entity_category=EntityCategory.DIAGNOSTIC
+    ),
```
GnoX, line 771: I do see why would you want to know the state of health, I'm just wondering whether it is not better suited as an attribute rather than a sensor. I'm not sure myself though, it's fine as a sensor, feel free to ignore.

GnoX:
> For AC output freq, it seems like it's not static.

I do understand that you can switch between 2 frequencies, I mean static in terms of once you switch it, it won't change until you hold the button for 10 seconds, you can't do it with a command.


Anyways, looks good.

GnoX (approved):

custom_components/ef_ble/sensor.py
MahdiBM: I am actually using this one. Though I don't have any preference on it being an attribute or sensor. Your call.

GnoX: Well, attributes don't contribute to the sensor clutter in the device view and since this changes very slowly, I thought it would be more appropriate to set it as an attribute of `battery_level`.  But I don't mind either way, sensor is fine.

rabits: Sorry, need all convos closed - @GnoX do you think it's resolved?

GnoX: My approval means that every was addressed so feel free to resolve it yourself if it's not already.

PR #423, opened 2026-07-19
Name circuit sensors and switches from the device, sorted and cached across restarts

GnoX:
This PR makes the circuit power and status sensors and the circuit on/off switches display device reported name name, as C{index} {name} {type} - so the entity reads "C28 Kitchen Power", "C28 Kitchen Status" and, for the switch, "C28 Kitchen". Because that name arrives only after the BLE link is up and can change at runtime, the name is composed by the integration from a {name}-bearing translation template rather than handed to Home Assistant's static name formatter, and the last-known value is cached across restarts.

Full list of changes:

- A shared DeviceNamedEntity mixin owns the whole behavior - reads the localized template, fills placeholders itself, persists via RestoreEntity; used by EcoflowNamedSensor/EcoflowNamedSwitchEntity, auto-selected for any description with a name_field.
- Circuit sensor/switch names became {name} templates in en/uk/zh-Hans, type word localized in place, C{index} leading; integration formats them (HA never sees the {name} slot).
- name_field plumbed through the controls system (for_each -> EntityType -> builder -> switch description).
- SHP2 now exposes circuit names via LoadChInfo.ch_name.
- Names survive a restart (cached per entity; live device name wins).

### DIFF
--- a/custom_components/ef_ble/description_builder.py
+++ b/custom_components/ef_ble/description_builder.py
@@ -28,6 +28,7 @@ class EntityDescriptionBuilder:
     _translation_key: str | None = None
     _translation_placeholders: dict[str, str] | None = None
     _availability_prop: str | None = None
+    _name_field: str | None = None
     _entity_registry_enabled_default: bool = True
 
     @classmethod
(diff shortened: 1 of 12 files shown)


GnoX:
@ecammit I remember we were discussing using circuit names reported by the device somewhere, this PR should do just that - can you check if this is working for your SHP2? [v0.44.0-a.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.44.0-a.1)

rabits (approved):
Yep looks good, tested on my SHP and it fetched the names.

PR #424, opened 2026-07-19
Expose SHP3 attached-battery SOC, power and serial as per-channel sensors

### LINKED ISSUE #412: [Sensor]: SHP3 per-channel battery SOC / power not exposed
### Pre-submission checks

### Device model

Smart Home Panel 3 (HR63)

### Sensor / control name

Channel battery SOC and power

HA exposes only the panel-wide battery level and power for the SHP3, while the SHP2 already has per-channel battery sensors. The SHP3's BLE stream carries each attached battery's serial, SOC, and signed power — visible in the attached diagnostics dump. One mapping note: the per-battery data identifies units by serial; I didn't find a channel-number field, so serial is the reliable key.

The batteries can also be added as their own BLE devices (I have both connected), but the panel stream carries this for the attached batteries over the single panel link, which would give per-battery data to setups that can't reach each unit directly.

### Integration version

1.0.1

### Diagnostics

[config_entry-ef_ble-01KW1KZZVB4C4PWB11X5SFR6EF-5.json](https://github.com/user-attachments/files/29938441/config_entry-ef_ble-01KW1KZZVB4C4PWB11X5SFR6EF-5.json)

### Anything else?

_No response_

GnoX:
This PR surfaces per-channel sensors that mirror the SHP2's (`channel{n}_battery_level`, `channel{n}_output_power`, `channel{n}_serial_number`), so a mixed SHP2/SHP3 setup reads consistently and setups that can only reach the panel still get per-battery data over the single panel link. The values come from the panel's own per-battery slots, tied to the backup channels through the panel's channel-to-battery pointer, so no proto changes were needed.

This is stacked on the circuit-naming PR (#423) and targets that branch; the two are independent in behavior but share the branch history.

Full list of changes:

- **The per-channel sensors are resolved through the panel's own indirection.** Each backup channel carries a device id that points at one of the panel's battery-info slots; computed field groups follow that pointer to expose `channel{n}_sn`, `channel{n}_battery_percentage` and `channel{n}_output_power`. Because those attribute names match the SHP2's existing sensor descriptions and translations, no new sensor entries or translation keys were needed - the SHP2's per-channel sensor definitions light up for the SHP3 as-is.
- **Panel-reported battery power is signed** (negative while the battery discharges into the panel); it is negated into `output_power` so discharge reads positive, matching the SHP2 convention.
- **The extra-battery slot scan was generalized.** The hardcoded 1-5 range in the sensor platform and config flow is replaced by an `extra_battery_indices()` helper that discovers the slots a device actually declares (up to 10), so the extra-battery harness no longer assumes a fixed count.
- **Attached batteries that are really full devices get their real model name.** A panel can report a standalone unit (e.g. a Delta Pro Ultra) as an attached battery; battery naming now falls back to the supported-device registry so it shows the actual model instead of a generic "Extra Battery". The registry lookup imports the devices package at module level rather than inside the function, to avoid a blocking import on the event loop.

Resolves #412

### DIFF
--- a/custom_components/ef_ble/config_flow.py
+++ b/custom_components/ef_ble/config_flow.py
@@ -61,7 +61,7 @@ from .const import (
     LINK_WIKI_SUPPORTING_NEW_DEVICES,
 )
 from .eflib.connection import Connection, ConnectionState
-from .eflib.device_mappings import battery_name_from_device
+from .eflib.device_mappings import battery_name_from_device, extra_battery_indices
 from .eflib.exceptions import AuthErrors
 from .eflib.logging_util import LogOptions
 from .eflib.login import EcoFlowLogin, Region
@@ -394,7 +394,7 @@ class EFBLEConfigFlow(ConfigFlow, domain=DOMAIN):
 
         if CONF_EXTRA_BATTERY not in entry_data:
             entry_data[CONF_EXTRA_BATTERY] = _find_enabled_batteries(
-                device, range(1, 6)
+                device, extra_battery_indices(device)
             )
 
         return self.async_create_entry(title=device.name, data=entry_data)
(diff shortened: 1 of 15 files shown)


rabits (approved):
Yep looks good!

PR #427, opened 2026-07-21
Validate GATT characteristics at connect time and clear stale service cache

### LINKED ISSUE #425: [Bug]:Device is using unsupported protocol for write
### Pre-submission checks

### What happened?

Errors in HA logs after HA restart:

CC:8D:A2:69:AF:00 = Delta Pro 3
Core = 2026.6.4
Supervisor = 2026.07.3
Operating System = 18.1
EcoFlow BLE = 1.0.2

Logger: custom_components.ef_ble.eflib.connection - CC:8D:A2:69:AF:00
Source: custom_components/ef_ble/eflib/connection.py:641
Integration: EcoFlow BLE (Unofficial) ([documentation](https://github.com/rabits/ha-ef-ble/wiki), [issues](https://github.com/rabits/ha-ef-ble/issues))
First occurred: 9:37:41 PM (1 occurrence)
Last logged: 9:37:41 PM

Captured exception: Device is using unsupported protocol for write. Available characteristics: : File "/config/custom_components/ef_ble/eflib/connection.py", line 815, in sendRequest await self._sendRequest(send_data, response_handler) File "/config/custom_components/ef_ble/eflib/connection.py", line 865, in _sendRequest await self._start_notify(response_handler) File "/config/custom_components/ef_ble/eflib/connection.py", line 852, in _start_notify await self._client.start_notify(self._notify_characteristic, callback, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.14/functools.py", line 1126, in __get__ val = self.func(instance) File "/config/custom_components/ef_ble/eflib/connection.py", line 708, in _notify_characteristic return self._get_characteristics("notify") ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^ File "/config/custom_components/ef_ble/eflib/connection.py", line 704, in _get_characteristics raise UnsupportedBluetoothProtocol("write", characteristic_list)

### Integration version

1.0.2

### Home Assistant version

2026.6.4

### Device model

Delta Pro 3

### Bluetooth connection

USB dongle

No

```text

```

### Diagnostics

_No response_

### Relevant log output

```shell

```

GnoX:
This PR makes the integration recover automatically when BlueZ hands it a connected client with a broken GATT service table, instead of failing with a captured `UnsupportedBluetoothProtocol` exception at the first send. The reported failure had an *empty* characteristics list right after an HA restart on a freshly updated host: BlueZ can report `ServicesResolved` against an empty or stale cache (typically after a bluetoothd restart or host update), and the integration only noticed at the first write, retried against the same dead service table, and never recovered without manual intervention.

Full list of changes:

- **Characteristics are validated at connect time.** Both the notify and write characteristics are resolved right after `establish_connection`, so a broken service table fails the connect itself, where HA's normal retry machinery (`ConfigEntryNotReady` with backoff, entry reload on disconnect) takes over, instead of surfacing deep inside the auth routine.
- **An empty service table triggers a GATT cache clear.** When validation fails and the device exposes *no* characteristics at all, the client's `clear_cache()` is called so the next attempt re-discovers services instead of resolving the same broken table again. The call is guarded with `getattr` since HA's client wrapper and `BleakClientWithServiceCache` provide it but plain `BleakClient` does not.
- `_notify_characteristic`/`_write_characteristic` are plain properties now; the previous `cached_property` was never invalidated, so a characteristic object resolved from an old client could leak into the next connection after a reconnect.
- `UnsupportedBluetoothProtocol` now names the characteristic type that actually failed to resolve (the reported traceback was a notify lookup misreported as "write") and exposes `characteristic_type`/`available_characteristics` for callers.
- Setup treats `UnsupportedBluetoothProtocol` like the other connect failures, so the entry shows "could not connect" with the real message instead of logging "Unknown error".

Resolves #425

### DIFF
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -46,7 +46,7 @@ from .eflib.connection import (
     ConnectionTimeout,
     MaxConnectionAttemptsReached,
 )
-from .eflib.exceptions import AuthErrors
+from .eflib.exceptions import AuthErrors, UnsupportedBluetoothProtocol
 from .eflib.logging_util import ConnectionLog
 
 PLATFORMS: list[Platform] = [
@@ -140,7 +140,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
         )
         async with asyncio.timeout(timeout):
             state = await device.wait_until_authenticated_or_error(raise_on_error=True)
-    except (ConnectionTimeout, BleakError, TimeoutError) as e:
+    except (
+        ConnectionTimeout,
+        BleakError,
+        TimeoutError,
+        UnsupportedBluetoothProtocol,
+    ) as e:
         await device.disconnect()
         raise ConfigEntryNotReady(
             translation_key="could_not_connect",
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -386,6 +386,15 @@ class Connection:
                 max_attempts=ble_attempts,
                 timeout=self._options.timeout,
             )
+            self._validate_characteristics()
+        except UnsupportedBluetoothProtocol as e:
+            error = e
+            if not e.available_characteristics:
+                # An empty service table is a host-side GATT cache glitch, not the
+                # device genuinely lacking the protocol - wipe the cache so the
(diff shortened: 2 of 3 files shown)


rabits (approved):
LGTM

PR #429, opened 2026-07-22
Move PowerStream to controls system and add feed-in control switch

### LINKED ISSUE #428: [Sensor]: Powerstream - feedProtect
### Pre-submission checks

Missing sensor / control (never appeared)

### Device model

Powerstream

### Sensor / control name

feedProtect

The sensor is used to change Powerstream behaviour, e.g. if battery is full it can control if all PV is still supplied to grid (all PV to grid) or only needed for household (PV limited to needed houesehold).

I can set it via HTTP / API by:
```
        elif cmd == "SET_FEEDPROTECT":
            if not (value == "0" or value == "1"):
                return {"unsupported value (0/1)": value}
            cmdCode = "WN511_SET_FEED_PROTECT_PACK"
            params = "feedProtect=" + value
            if DEBUG: LOGGER.debug(params)
            body = {
                "sn": sn,
                "cmdCode": cmdCode,
                "params": {
                    "feedProtectPriority": value
                }
            }
```

Ecoflow Documentation:
https://developer-eu.ecoflow.com/us/document/PP3?id=2058828968760086530
feedProtect | int | Feed-in control: 0: off; 1: on

### Integration version

1.0.2

### Diagnostics

if diagnostics needed, I need to learn to dump first ...

Maybe you have existing logs from previous analysis... 

### Anything else?

_No response_

GnoX:
This PR migrates the PowerStream's existing controls to the `@controls` decorator system and adds a switch for feed-in control (`feedProtect`), which decides whether surplus PV power may be exported to the grid.

Full list of changes:

- **The `load_power` number and `power_supply_priority` select move from the deprecated hardcoded descriptions to `@controls` decorators** on the device. Entity keys are unchanged, so existing entities carry over; the load power limit still comes from the device's `load_power_max`.
- **New `feed_protect` switch.** The device already had the write method; it is now backed by a state field read from the heartbeat and exposed as a switch.

Resolves #428


rabits (approved):
Looks great!

PR #431, opened 2026-07-22
Remove charge limit override in River 3 set_ac_charging_speed

### LINKED ISSUE #430: [Bug]: Setting AC charging speed on River 3 (Plus) resets max charge limit to 12%
### Pre-submission checks

### What happened?

Steps to reproduce:
1. Set `number.…_battery_charge_limit_max` to any value (e.g. 85%)
2. Change `number.…_ac_charging_speed` to any value
3. Read back the max charge limit

Expected: only the AC charging speed changes.
Actual: the maximum battery charge limit is silently reset to 12%.

Root cause is in `river3.py` — `set_ac_charging_speed` begins with a hardcoded `await self.set_battery_charge_limit_max(12)` before the actual charging-speed write:
https://github.com/rabits/ha-ef-ble/blob/master/custom_components/ef_ble/eflib/devices/river3.py#L257

The call was introduced in the #249 control-registration refactor — the pre-refactor implementation of `set_ac_charging_speed` had no such call, no other device module has one, and it's unrelated to the charging-speed config write that follows. It looks like a debugging call that was accidentally left in. Removing it fixes the behaviour on my River 3 Plus with no side effects (verified: changing the charging speed leaves the configured limit untouched).

PR with the fix and a regression test: #431

### Integration version

1.0.2

### Home Assistant version

2026.7.2

### Device model

River 3 Plus (firmware version not exposed via BLE)

### Bluetooth connection

USB dongle

No

### Diagnostics

Not attached — the cause is identified at code level (link above). Happy to provide a dump if useful.

### Relevant log output

```shell
None — no errors are logged; the limit change happens silently.
```

melissamcdougall:
Fixes #430

`set_ac_charging_speed` in `river3.py` starts with `await self.set_battery_charge_limit_max(12)`, which forces the max charge limit to 12% every time the AC charging speed is changed - discarding whatever limit the user had configured. The call was introduced in the #249 refactor (the pre-refactor implementation didn't have it), no other device module does this, and it's unrelated to the `cfg_plug_in_info_ac_in_chg_pow_max` write below it.

This PR removes the call.

Verified on a River 3 Plus (HA 2026.7.2, USB Bluetooth dongle).

- `uv run pytest tests/eflib` — 154 passed
- `uvx prek run --all-files` — all hooks pass

### DIFF
--- a/custom_components/ef_ble/eflib/devices/river3.py
+++ b/custom_components/ef_ble/eflib/devices/river3.py
@@ -254,8 +254,6 @@ class Device(DeviceBase, ProtobufProps):
         max=dynamic(max_ac_charging_power),
     )
     async def set_ac_charging_speed(self, value: float):
-        await self.set_battery_charge_limit_max(12)
-
         if (
             self.max_ac_charging_power is None
             or value > self.max_ac_charging_power


GnoX:
What the heck, how did this happen. I completely missed this.

tests/eflib/test_river3.py
```diff
@@ -223,3 +224,18 @@ async def test_river3_exact_values_from_known_packets(device, packet_sequence):
...
+
+    result = await device.set_ac_charging_speed(200)
+
+    assert result is True
+    device._conn.sendPacket.assert_awaited_once()
+    packet = device._conn.sendPacket.await_args.args[0]
+    config = pr705_pb2.ConfigWrite()
+    config.ParseFromString(packet.payload)
+    assert config.cfg_plug_in_info_ac_in_chg_pow_max == 200
+    assert not config.HasField("cfg_max_chg_soc")
```
GnoX: Testing for this is just silly, this is not going to happen again. Just remove this whole test.

GnoX (changes requested):

GnoX (approved):

rabits (approved):
Nice catch! Thank you

PR #441, opened 2026-07-29
Add Power Kits Power Hub support

### LINKED ISSUE #440: [Device]: Power Kits Power Hub
## Pre-submission checks

- [x] The device is not listed in the Supported Devices section of the README
- [x] I searched open and closed issues and found no existing request for this device
- [x] I read the new-device support and diagnostics guidance

## Device model

EcoFlow Power Kits Power Hub

## Firmware version

Not confirmed yet.

## Serial number prefix

`M109`

## Bluetooth connection tested

Internal Bluetooth adapter on macOS.

## Diagnostics collection status

The device connects and authenticates over local BLE. A 100-packet diagnostic was collected locally while developing the parser, but that local capture was intentionally generated without encryption and will not be uploaded. I will add a fresh encrypted capture, preferably with non-zero PV input, before the linked PR is marked ready for review.

## Features requested

- Total battery state of charge
- Total input and output power
- Battery bus voltage and current
- PV1 and PV2 voltage, current, power, and temperature
- Charge and discharge time remaining
- Up to three connected Power Kits LFP battery packs: state of charge, temperature, voltage, cell-voltage range, input power, and output power

## Hardware validation

Initial support is running locally on Home Assistant 2026.7.4. A night-time capture produced stable values for the Power Hub and all three attached `M101` battery packs. The locally observed aggregate values included 73% state of charge, 53.166 V bus voltage, and pack state-of-charge values of 79%, 71%, and 71%.

PV field offsets and scaling have been cross-checked against the MIT-licensed Power Kit CAN research in GridSpace/wattzup, but non-zero PV values have not yet been verified on this physical device. The implementation will therefore start as a draft PR.

## Draft PR

Will be linked here after it is opened.

jkqq147:
## Summary

- add local BLE discovery and device mapping for the `M109` Power Kits Power Hub
- decode aggregate BMS, MPPT, and per-pack records with the existing fixed-width `RawData` patterns
- expose total battery/power data, two PV inputs, controller temperatures, and detailed data for up to three `M101` LFP battery packs
- use the app-declared signed 32-bit field widths, distinguish pack/cell/heatsink temperatures, and expire battery slots not seen in a new authenticated session
- add English, Simplified Chinese, and Ukrainian names plus masked real-device parser coverage

Closes #440.

## Why

The Power Kits Power Hub currently falls back to diagnostic-only unsupported-device mode even though it provides authenticated local BLE telemetry. Supporting it directly provides a read-only, fully local path without EcoFlow cloud/MQTT telemetry.

## Protocol basis

The Power Hub uses v3/`0x13` packets with the existing sequence-byte payload XOR convention:

- `src=0x03`, `cmd_set=0x03`, `cmd_id=0x1c`: aggregate BMS record
- `src=0x03`, `cmd_set=0x03`, `cmd_id=0x1a`, `dsrc=1..3`: one attached battery record per slot
- `src=0x05`, `cmd_set=0x05`, `cmd_id=0x20`: MPPT/solar-controller record

The record layouts were reconstructed from a local BLE diagnostic and checked against the field order/types in EcoFlow's named Power Kits app records (`BmsTotalDataBean`, `BmsDataBean`, and `SccDataBean`). The resulting offsets also agree with GridSpace/wattzup's MIT-licensed Power Kit CAN research:
https://github.com/GridSpace/wattzup/blob/main/lib/can/structs.json

## Hardware validation

Tested on a physical `M109` Power Hub with three attached `M101` batteries, using Home Assistant 2026.7.4 and the Mac's internal Bluetooth adapter.

Night-time capture:

- the masked captured pack record decodes `current=-61` (signed, milliamp scale), `voltage=53194`, SOC 71%, and 30/29 °C max/min cell temperatures
- aggregate SOC/output decode as 73% / 7 W
- both PV inputs correctly decode as 0 W

Daylight live validation:

- an aligned database sample reported PV1 as 35.833 V, 4.613 A, and 165 W; `V × A = 165.3 W`
- PV2 remained at 0 V / 0 A / 0 W
- all three battery devices remained live with SOC, voltage, pack temperature, and per-pack power
- a separate encrypted diagnostic collected 1,000 raw messages for upstream-safe troubleshooting

The reviewer-proposed 20-second RTC keepalive was tested exactly as suggested. It did not change the device behavior: three sessions still disconnected about 30.6 seconds after authentication. The shared entity-preserving reconnect implementation has therefore been removed from this PR; reconnect behavior can be handled as a focused follow-up after the base device support lands.

## Scope and known limitations

- read-only sensors only; no controls
- discovery is restricted to the observed `M109` prefix
- remaining-time entities are deliberately withheld until their wire unit can be compared with the EcoFlow app
- existing config entries created in unsupported mode may need reconfiguration to select the detected extra batteries
- reconnect handling is intentionally outside this device-support PR

## Tests

- `uv run pytest` — 161 passed (15 pre-existing SHP3 `AsyncMock` warnings)
- `uvx prek run --all-files` — all hooks passed
- Home Assistant configuration check and restart on the physical-device host
- live dashboard verification across a device-initiated disconnect/reconnect cycle

## AI assistance

This is an AI-assisted new-device draft. Codex was used to analyze the local diagnostics, implement the repository-native fixed-width model, and prepare tests/documentation. Claims above are limited to the app schema, masked captured records, and live hardware observations.


GnoX:
> [!NOTE]

I checked your three record layouts against the EcoFlow app, which carries the Power Kits records as named field lists (`BmsTotalDataBean`, `BmsDataBean`, `SccDataBean` - the last two even keep `dSrc`/`dDst`, so they are these BLE records, not a cloud shape). Reconstructing each record from the app's field order and declared types reproduces **every one of your 21 offsets exactly**, including the PV ones you flagged as unvalidated. So the hard part is right, and PV positions no longer need daylight to be trusted - only the scaling does.

What the comparison does turn up is that the field *types* are narrower than the wire, and two sensors are named after the wrong field.

**`battery_current` is signed, and is probably wrong in your install right now.** The app declares `batteryCur` as a 4-byte signed int; you read it as `H`. A discharging hub gives a negative value, so a 2-byte unsigned read of, say, -1.69 A returns `0xF966` = 63846 -> **63.846 A**. Your night-time capture was discharging, so this is worth looking at before anything else. It should be `i`.

**The same narrowing runs through the rest.** In the app all of these are 4-byte ints: pack `vol`, `maxCellVol`, `minCellVol`, and in the SCC record `batteryVol`, `pv1InCur`/`pv2InCur`, `pv1InWatt`/`pv2InWatt`, plus the aggregate `totalInputWatt`/`totalOutputWatt`. Reading them as `H`/`h` happens to work while the value fits in 16 bits and stays non-negative, which is why your night capture looked clean - `pv_voltage_1` is the one you already made `I`. Widening them is mechanical and removes a whole class of surprise (a pack that exports, an output over 32767 W).

**`pv_temperature_1` / `pv_temperature_2` are heatsink temperatures**, `hs1Temp` and `hs2Temp`, declared as signed 2-byte shorts - and `pcbTemp` follows at offset 76. Read as `B` you get the low byte only, so -5 °C reports 251. `h` fixes it, and the names are worth revisiting.

**`cell_temperature` per pack is the pack temperature.** Offset 46 is the app's `temp`; the actual cell temperatures are `maxCellTemp` at 73 and `minCellTemp` at 74, both single bytes. The integration already distinguishes these two things elsewhere (`battery_temperature` vs `cell_temperature`), so this one is just a rename plus two fields you can now expose for free.

**`reserved_1` in the aggregate record is `totalChgDsgState`** - the device's own charge/discharge direction flag, sitting right where you need it. That replaces deriving direction from `input_power` vs `output_power`, which is fragile at low idle loads (and as written, `input_power == output_power` yields discharge but never charge). While you're there: the app has a single `totalRemainTime` like you do, but every other device in this repo passes its remaining-time field straight into a `UnitOfTime.MINUTES` sensor (`_delta2_base.py:93-94`, `sensor.py:245`), so dividing by 60 assumes seconds against the family convention. Worth one comparison with what the app displays.

**Your reserved gaps are all named fields**, if you want them later: pack `amp@42`, `chargeState@47`, `openBmsState@48`, `designCapacity@49`, `fullCapacity@53`, `remainCapacity@57`, `recycles@61`, mos temps offset 75/76, `bmsFault@77`, `remainTime@86`, and `chgSetSoc`/`dsgSetSoc` (the obvious basis for controls later); SCC `batteryWatt@36`, `l1Cur@64`, `l2Cur@68`, plus `dayEnergy`/`totalEnergy` further in, which would feed the energy dashboard; aggregate `bmsChgStopSoc@15` and `bmsDsgStopSoc@19`.

Two things unrelated to the layout:

- **The `eflib/__init__.py` guard is dead code.** `unsupported.py` defines `UnsupportedDevice` but no module-level `Device`, so `getattr(item, "Device", None)` is already `None` there and the loop never considers it. Reverting that hunk on your branch leaves all 158 tests passing, including your own `test_new_device_selects_power_hub_before_unsupported` - so that test passes with or without the change. Dropping it leaves the PR with no changes to shared device selection, which makes it much easier to merge. If your hub really was landing in unsupported mode, the cause is the pre-existing config entry you mention under limitations.
- **The discovery matcher is wider than the supported range.** `manufacturer_data_start: [18, 77, 49, 48]` is `0x12` + `"M10"`, so it matches any `M10x` while `SN_PREFIX` claims only `M109`. If the `M101` packs advertise on their own, Home Assistant will start offering them as unsupported devices. Adding `57` (`"9"`) keeps discovery aligned with what the PR actually supports - you'd know better than me whether the packs advertise at all.
- **`battery_enabled` never goes back to `False`.** It is set `True` whenever a record arrives, so a detached or sleeping pack keeps its last SOC and voltages indefinitely. The existing extra-battery devices drive that flag from the device's availability flag instead (`_delta2_base.py:237`, `delta_pro.py:243`).

On tests: the current ones pack each value at the same offset the model reads, so they pass whatever the offsets and widths are - the `battery_current` issue above survives them untouched. The repo convention for raw-data devices is a captured frame as the fixture (see the `_test_*_packets.py` stubs), which would have caught it. Your masked encrypted capture would make a good fixture when you add it.

Also minor: `lambda value: max(0, value)` is the inline-lambda case `CLAUDE.md` asks to avoid (`props/transforms.py` has `out_power`), and clamping at zero hides reverse flow; the two bare `/1000` divisions in `_update_battery` could use `pdiv(1000, 3)` like the rest of the file.

Good PR - the read-only-first scope and the draft status are the right shape. Merge decisions are GnoX's when he's back on Aug 8; this is review to save you a round trip.

---
🤖 *AI-generated review (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). A human will review this properly after Aug 8.*

jkqq147:
Added the hardware-validated Power Hub reconnect refinement in `c73291e`.

The physical `M109` still closes the CoreBluetooth connection about 31 seconds after authentication. This update deliberately does not guess at an undocumented keepalive command. Instead, only Power Hub opts into a fresh BLE-session reconnect while its existing Home Assistant entities remain loaded; failure falls back to the previous full config-entry reload.

Four consecutive device-initiated disconnects recovered to the first authenticated packet in 4.420 s, 4.273 s, 4.310 s, and 4.428 s. During a separate 45-second dashboard sample spanning a reconnect, SOC and the local-BLE indicator remained present in all 45 samples, without a Power Hub offline/waiting state.

Validation:

- `uv run pytest tests/eflib` — 159 passed
- `uvx prek run --all-files` — all hooks passed
- Home Assistant config check/restart and live hardware validation on HA 2026.7.4

custom_components/ef_ble/eflib/devices/power_hub.py
```diff
@@ -0,0 +1,145 @@
...
+bms = dataclass_attr_mapper(PowerHubBmsData)
+mppt = dataclass_attr_mapper(PowerHubMpptData)
+
+
+class Device(DeviceBase, RawDataProps):
+    """Power Kits Power Hub"""
+
+    SN_PREFIX = (b"M109",)
+    NAME_PREFIX = "EF-M10"
+    RECONNECT_IN_PLACE = True
```
GnoX: This is where I'd try the keepalive first. `SHP3` holds an otherwise-dropping link with an RTC check every 20 s (`shp3.py:148`, `shp3.py:346`, `shp3.py:353`), and the hub already answers `RET_TIME`, so the command is not new surface:

```suggestion
    _KEEPALIVE_INTERVAL = 20
```

with `self.add_timer_task(self._send_keepalive, interval=self._KEEPALIVE_INTERVAL)` in `__init__` and `_send_keepalive` calling `self._time_commands.sendRTCCheck()`. If the hub stops dropping at 31 s, both reconnect constants and the whole `__init__.py` change come out.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

GnoX (commented):
The reconnect works, and the timings you measured show it works well. But the 31-second drop is very likely a missing keepalive rather than a property of this device, and this repo already has the command for it - so I'd try that before adding a reconnect path to the shared integration.

You wrote that you did not want to guess at an undocumented keepalive, which is the right instinct. You don't have to guess: `SHP3` faces the same idle-disconnect behaviour and holds its link with a plain RTC check every 20 seconds (`shp3.py:148` and `shp3.py:346,353`), using `TimeCommands.sendRTCCheck()` - a command already in the library. Your own `data_parse` answers the hub's `RET_TIME` request with `async_send_all()`, so the hub demonstrably speaks this exact time-sync exchange; sending it periodically is not new protocol surface.

If that holds the link, everything in this commit disappears: the 91 lines in `custom_components/ef_ble/__init__.py`, both class constants in `devicebase.py`, and the reconnect test. It also removes the cost the current approach pays - a full key exchange and authentication every ~31 seconds, and a 4.3 s gap in every cycle, so roughly 13% of the time with no data and a busy radio on adapters that are often shared with other devices.

Worth trying as a two-line experiment on your hardware before anything else:

```python
_KEEPALIVE_INTERVAL = 20

# in __init__, after self._time_commands = TimeCommands(self)
self.add_timer_task(self._send_keepalive, interval=self._KEEPALIVE_INTERVAL)

async def _send_keepalive(self) -> None:
    await self._time_commands.sendRTCCheck()
```

If the hub still drops at 31 seconds with that running, then the reconnect path has a real justification and this becomes a much easier review - please say so either way, since "keepalive does not help this device" is itself worth recording.

The layout findings from my previous comment are unaffected by this commit and still open - `battery_current` reading ~63.8 A while discharging is the one I would fix first.

---
🤖 *AI-generated review (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). A human will review this properly after Aug 8.*

custom_components/ef_ble/eflib/devices/power_hub.py
```diff
@@ -0,0 +1,145 @@
...
+    RECONNECT_DELAY = 1.0
```
GnoX: `RECONNECT_DELAY = 1.0` is identical to the `DeviceBase` default this commit adds, so the override has no effect - worth dropping either way.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

custom_components/ef_ble/__init__.py
```diff
@@ -69,6 +69,85 @@
...
+    await device.connect(user_id=user_id, max_attempts=2)
+    async with asyncio.timeout(timeout):
+        reconnect_state = await device.wait_until_authenticated_or_error(
+            raise_on_error=True
+        )
+    if not reconnect_state.authenticated:
+        raise ConnectionError(f"Reconnect ended in state {reconnect_state}")
+
+
+class _ReconnectManager:
```
GnoX: This is the part that makes the PR much harder to merge: a per-device reconnect strategy in the integration's shared setup path, added to work around one device's idle timeout. Each cycle pays a full ECDH exchange plus authentication and leaves a ~4.3 s hole in the data, about 13% of the time on a radio that is often shared with other devices. If the 20 s RTC keepalive holds the link, none of this is needed; if it genuinely doesn't, this belongs in its own PR with the reasoning attached, so the device support can merge on its own.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

custom_components/ef_ble/__init__.py
```diff
@@ -69,6 +69,85 @@
...
+    ) -> None:
+        self._hass = hass
+        self._entry = entry
+        self._device = device
+        self._address = address
+        self._user_id = user_id
+        self._timeout = timeout
+        self._task: asyncio.Task[None] | None = None
+
+    def on_disconnect(self, _exc: Exception | type[Exception] | None) -> None:
```
GnoX: One behavioural note if this path stays: on a device that has really gone away, `_async_reconnect_in_place` burns `RECONNECT_DELAY` + `max_attempts=2` + the auth timeout before falling back to `async_schedule_reload`, so recovery is slower than today's immediate reload for the common case of the unit being switched off. A cheap guard is to skip the in-place attempt when `async_last_service_info` shows no recent advertisement.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

tests/eflib/test_power_hub.py
```diff
@@ -0,0 +1,105 @@
...
+def test_new_device_selects_power_hub_before_unsupported(mocker: MockerFixture):
+    ble_dev = mocker.Mock()
+    ble_dev.address = "AA:BB:CC:DD:EE:FF"
+    adv_data = mocker.MagicMock()
+    adv_data.manufacturer_data = {0xB5B5: b"\x12M109TEST0000001"}
+
+    assert isinstance(eflib.NewDevice(ble_dev, adv_data), Device)
+
+
+def test_power_hub_uses_fast_in_place_reconnect():
```
GnoX: This asserts the constants are the values they are declared as, so it passes no matter how the reconnect behaves. If the path stays, the behaviour worth pinning is `_ReconnectManager.on_disconnect`: a device without `RECONNECT_IN_PLACE` schedules a reload, one with it starts exactly one task, and a second disconnect while that task runs does not start another.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

custom_components/ef_ble/eflib/devices/power_hub.py
jkqq147: Removed from the upstream device-support diff. Reconnect behavior is now outside this PR.

tests/eflib/test_power_hub.py
jkqq147: Removed the constants-only test together with the reconnect implementation. The remaining Power Hub tests now cover signed field widths, masked captured payloads, temperature semantics, and unseen battery-slot expiry. A future reconnect PR will test manager behavior rather than declarations.

custom_components/ef_ble/__init__.py
jkqq147: This path has been removed from the PR, so the stale-advertisement/fallback concern no longer applies here. A future reconnect PR will include an advertisement freshness guard and explicit fallback behavior.

custom_components/ef_ble/eflib/devices/power_hub.py
jkqq147: Tested this exact 20 s RTC-check keepalive on the physical M109. It still disconnected 30.76 s, 30.57 s, and 30.60 s after authentication, so the keepalive experiment has been removed. I also removed the shared reconnect implementation from this PR and documented the result in the updated description.

custom_components/ef_ble/__init__.py
jkqq147: Agreed. The shared reconnect manager and DeviceBase opt-in have been removed from this PR. Since the RTC check did not hold the physical link, reconnect behavior will be handled as a focused follow-up after base M109 support lands.

jkqq147:
Addressed the automated review in `552da6a` and `cfb5d95`:

- widened the app-declared current/voltage/power fields to signed 32-bit values
- corrected pack temperature vs min/max cell temperature and PV heatsink temperature naming
- added PCB temperature and masked captured-night payload coverage
- restricted discovery to `M109`
- removed the dead `NewDevice` guard
- added authenticated-session battery-slot expiry so detached packs do not retain stale measurements
- removed the inline output clamps and used shared scaling transforms
- withheld remaining-time entities until the unit is verified against the EcoFlow app
- tested the suggested RTC keepalive on hardware; it did not stop the ~31 s disconnect
- removed the entire shared reconnect change and its constants-only test from this PR

Fresh validation:

- `uv run pytest` — 160 passed
- `uvx prek run --all-files` — all hooks passed
- daylight sample: 35.833 V × 4.613 A = 165.3 W vs 165 W reported
- Home Assistant config check, restart, three-pack dashboard, and one reconnect cycle verified on the physical M109

The PR remains Draft for the upstream human review.

jkqq147:
Final cleanup in `cfa6497`: the per-pack field-group container names no longer collide with the main-device `battery_input_power` / `battery_output_power` sensor keys. Public per-pack entity fields remain unchanged, and a regression test now guards against accidental main-device sensor creation.

Fresh upstream-branch validation:

- `uv run pytest` — 161 passed
- `uvx prek run --all-files` — all hooks passed
- Home Assistant configuration check and physical M109 verification with three live M101 battery packs

The Power Hub-specific reconnect refinement remains outside this PR on the fork deployment branch, so this PR stays limited to clean read-only device support.

GnoX:
Verified against the app's own record definitions: your corrected layout now matches them field for field, with every one of the gaps I listed filled at the right offset and width - pack `voltage@38` and `current@42` as signed 32-bit, `battery_temperature@46`, capacities at 49/53/57, `cycle_count@61`, cell voltages at 65/69, cell and mos temps at 73-76, `bms_fault@77`, powers at 78/82, `remaining_time@86`, and on the solar side `battery_power@36` recovered plus `pcb_temperature@76` and the signed heatsink shorts. 161 tests pass here on `cfa6497`, and no shared file is touched any more.

The daylight sample settles the scaling question too: 35.833 V x 4.613 A = 165.3 W against 165 W reported is exactly the agreement you would expect if the fields are mV, mA and W, which is what the app's declarations imply.

One field left over, and it is latent rather than broken: the aggregate `remaining_time` is still `H` while the app declares that record's remaining-time as a 32-bit int, like the per-pack one you widened. Nothing reads it while the entities are deferred, so it only matters when you come back to them.

Thanks for running the keepalive test rather than taking my word for it - a negative result on hardware is worth more than my inference from `SHP3`, and it is the right basis for #442 existing at all. I have left review notes there.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

jkqq147:
Closing this proposal because we have decided to maintain the Power Hub integration in the jkqq147/ha-ef-ble fork rather than pursue upstream inclusion. Thank you for the protocol review and feedback; the branch and validated implementation remain preserved in the fork.

GnoX:
Understood, and thanks for saying so plainly rather than letting it go quiet.

One thing worth stating in case it factored in: none of my review was a gate. GnoX is away until August 8 and merge decisions were always his, so a pile of review notes from an agent working in his repo is not the same as upstream declining the work. If you would rather it live in your fork regardless, that is a perfectly reasonable call for a device only you can test - but if the only reason to close was the review traffic, the door is open and he can decide when he is back.

Either way the protocol work stands on its own. The record layouts, the field widths and semantics, the ~30.6 s post-authentication session limit and the negative keepalive result are all written down in these two threads now, which is more than existed for Power Kits anywhere before this week. Anyone who comes looking - including whoever eventually adds M109 upstream - starts from your findings rather than from a blank capture.

Thanks for the care you put into it, and for testing the keepalive on hardware instead of taking my inference at face value.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

GnoX:
Back from vacation and caught up on the thread. The final layout matches the app's record definitions field for field and it's verified on your hardware - this is solid work, and I'd rather not see Power Kits support end up in a fork people have to find first.

Which is why I'd like to know what made you close it. If it was the review traffic while I was away -those notes were input for my merge decision, not a verdict, and I never got to make one.

I could take over the implementation and maintain it here, based on your branch  (the reconnect topic from the follow-up PR I'll handle separately). All I'd need from you is testing, since you're the one with the M109 and live packs.

Two asks, if you're up for it:

1. A diagnostics dump now, captured while the hub is charging or under a known load, with what the app showed at that moment (SOC, watts, temps, remaining time). Real frames give me proper test fixtures and settle the remaining-time unit. On the privacy doubts you mentioned: the dump is encrypted against the developer key (on by default, just don't disable the encrypt option), so to anyone else downloading the attachment it's ciphertext - only I can decrypt and read it.
2. Try a build once I have one ready and tell me whether the numbers match reality.

And if you'd rather keep it in your fork after all, just say so and I'll leave it be.

Jethro1999-gif (commented):
All right, I love what I'm seeing here! Do I need to replace and add each file with the file editor, or is there a complete repository link I can use for this fork?

PR #442, opened 2026-07-30
Preserve Power Hub entities across BLE reconnects

jkqq147:
## Dependency

Depends on #441. This PR is intentionally Draft until the base M109 device support lands. After #441 is merged, the branch will be rebased onto the latest `main` so the review contains only the reconnect delta.

## Summary

- let devices opt into entity-preserving reconnect while retaining the existing full config-entry reload as the default
- reconnect the Power Hub with a fresh BLE session without unloading its Home Assistant entities
- keep the last valid values available only while the in-place reconnect is active
- require a recent connectable advertisement before attempting the in-place reconnect
- run at most one reconnect task, cancel it during config-entry unload, and fall back to the original reload behavior on any failure
- add behavior tests for the default path, duplicate disconnects, reconnect failure, unload cancellation, and transient availability

## Why

The physical M109 used to validate #441 closes its CoreBluetooth connection about 31 seconds after authentication. The existing 20-second `sendRTCCheck()` keepalive was tested on the hardware and did not change the disconnect interval.

The integration currently responds to a disconnect by reloading the full config entry. That recovers eventually, but it also unloads and recreates the entity layer. This change keeps the entity objects and Dashboard state loaded while replacing only the expired BLE session.

Longer hardware validation found an additional timing issue: entity availability previously depended directly on `device.is_connected`. If a Home Assistant state write landed inside the roughly four-second reconnect window, the entire Power Hub entity set became `unavailable` until the next throttled write about 30 seconds later. The reconnect manager now exposes a narrowly scoped transient state so entities retain their last valid values during that window.

Only devices that explicitly set `RECONNECT_IN_PLACE = True` use the new path. All existing devices preserve the current immediate-reload behavior.

## Failure behavior

The reconnect attempt is deliberately bounded:

- wait one second after disposing of the old connection
- require an advertisement no older than 60 seconds
- use the existing two-attempt connection flow and authentication timeout
- clear transient reconnect availability on success, failure, or cancellation
- schedule the original config-entry reload if any reconnect step fails

## Hardware validation

Validated on a physical M109 Power Hub with three attached M101 batteries, Home Assistant 2026.7.4, and the Mac's internal Bluetooth adapter.

A 16-minute post-restart soak covered 30 device-initiated disconnects and 30 successful authentications. Reauthentication took approximately 4.2-4.6 seconds. SQLite history recorded zero new `unknown` or `unavailable` states for the Dashboard's `sensor.ef_m100054_*` entities, including samples taken while a reconnect was in progress.

PV1 also changed during the soak, confirming that retained values were not frozen and fresh packets continued updating entities after reconnect. The three battery devices remained present and continued exposing SOC, voltage, temperature, and per-pack power without a full config-entry reload.

## Tests

- `uv run pytest tests/eflib` - 166 passed (15 pre-existing SHP3 `AsyncMock` warnings)
- `uvx prek run --all-files` - all hooks passed
- Home Assistant configuration check and restart on the physical-device host
- 16-minute live Dashboard/database verification across 30 disconnect/reconnect cycles

## AI assistance

This is an AI-assisted follow-up. Codex was used to isolate the reconnect coordinator, add behavior tests, and validate the result on the physical device. Claims above are limited to the observed M109 behavior and the recorded test/runtime evidence.


custom_components/ef_ble/__init__.py
```diff
@@ -67,6 +69,36 @@
 ConfigEntryError = partial(ConfigEntryError, translation_domain=DOMAIN)
 
 _REAPPEAR_CALLBACKS_KEY = f"{DOMAIN}_reappear_callbacks"
+_IN_PLACE_RECONNECT_DELAY = 1.0
```
GnoX, line 72: Nothing limits how often this can cycle. With a device that drops ~31 s after auth this is the intended steady state, but consider the failure case: if the hub disconnects immediately after authenticating - wrong user id, another client stealing the link, a firmware sulk - `on_disconnect` fires again straight away and the only brake is the 1 s delay. The entry never reloads, so the user sees entities that exist but never update, and `max_connection_attempts_reached` (raised on the reload path) never appears.

Counting consecutive short-lived sessions and giving up after a few would keep the good case fast and let the existing backoff/repair machinery handle the bad one:

```suggestion
_IN_PLACE_RECONNECT_DELAY = 1.0
_MAX_RECONNECT_ADVERTISEMENT_AGE = 60.0
# A session shorter than this counts as a failed reconnect, not a working one
_MIN_HEALTHY_SESSION = 10.0
# Consecutive short sessions before handing back to the reload path and its repair issue
_MAX_RAPID_RECONNECTS = 3
```

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

GnoX (commented):
> [!NOTE]

Splitting this out was the right call, and the shape is much better than the version inside #441: the coordination logic is its own module with real behaviour tests, the advertisement-freshness check is in, and the reload fallback is preserved. 165 tests pass here on `a908794`.

The keepalive result is the part that matters most, and it changes my earlier objection: with a 20-second `sendRTCCheck()` making no difference on hardware, this is not an idle timeout being papered over, so an entity-preserving reconnect is a reasonable answer rather than a workaround. Please keep that sentence in the PR body - it is the evidence that justifies the whole change, and without it a future reader will ask the same question I did.

One substantive gap, noted on the diff: nothing bounds how often this can cycle. On your hardware a ~31 s session followed by a ~4 s recovery is the expected steady state, but a device that drops immediately after authenticating turns the same path into a tight loop that never surfaces a failure, and the in-place path also skips the `max_connection_attempts_reached` repair issue that the reload path raises today. A cheap guard is to count consecutive sessions that ended sooner than some threshold and hand back to the reload path after a few.

Two smaller things are on the diff as well, plus a hypothesis for what the 31 seconds actually is, which is worth one hardware test before this is treated as permanent.

Since this depends on #441, none of it is urgent - and merge order and merge decisions are both GnoX's when he is back on Aug 8.

---
🤖 *AI-generated review (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). A human will review this properly after Aug 8.*

custom_components/ef_ble/eflib/reconnect.py
```diff
@@ -0,0 +1,57 @@
...
+        if self._task is not None and not self._task.done():
+            return
+        self._task = self._create_task(self._run())
+
+    async def _run(self) -> None:
+        try:
+            await self._reconnect()
+        except asyncio.CancelledError:
+            raise
+        except Exception as exc:  # noqa: BLE001
```
GnoX, line 45: House rule in `.claude/rules/code-style-rules.md`: every suppression carries a short reason on the same line. The catch itself is justified here - any failure has to reach the fallback - so it is only the annotation that is missing.

```suggestion
        except Exception as exc:  # noqa: BLE001 - any failure must fall back to a reload
```

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

custom_components/ef_ble/eflib/devices/power_hub.py
```diff
@@ -0,0 +1,168 @@
...
+mppt = dataclass_attr_mapper(PowerHubMpptData)
+millivolts_to_volts = pdiv(1000, 3)
+
+
+class Device(DeviceBase, RawDataProps):
+    """Power Kits Power Hub"""
+
+    SN_PREFIX = (b"M109",)
+    NAME_PREFIX = "EF-M10"
+    RECONNECT_IN_PLACE = True
```
GnoX, line 25: With the RTC keepalive ruled out, one hypothesis is still untested and it would be better than reconnecting forever: several EcoFlow families drop a client that authenticates but never registers a user id. `shp3.py` sends exactly such a packet after auth (`_send_userid_registration`, `0x21 -> 0x35`, `cmd_set 0x35`, `cmd_id 0xA8`, user id padded to 64 bytes plus a little-endian timestamp) and OCEAN Pro does the same. If the hub is enforcing an unregistered-client timeout rather than an idle one, that would explain why traffic alone does not help.

Also useful to know either way: is the ~31 s measured from connection or from authentication, and is it constant across sessions? A fixed window from connect looks like a session lifetime you cannot avoid; a fixed window from auth points at something expected shortly after auth. Worth a line in the code comment once known, so nobody re-litigates this.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies on this thread.*

jkqq147:
Closing this follow-up because we have decided to maintain the Power Hub reconnect behavior in the jkqq147/ha-ef-ble fork rather than pursue upstream inclusion. Thank you for the review and failure-mode feedback; the branch and validated implementation remain preserved in the fork.

PR #450, opened 2026-08-07
Improve STREAM multi-device BLE reconnect and telemetry

MrChurch:
## Summary

This PR improves BLE stability for multiple EcoFlow STREAM AC and STREAM AC Pro devices.

### Changes

- Fixed reconnect handling when the connection enters the `RECONNECTING` state.
- Added per-device reconnect backoff and jitter to avoid simultaneous reconnect storms.
- Explicitly enabled STREAM property telemetry after authentication.
- Reapplied the telemetry configuration after reconnects.
- Added regression tests for telemetry activation and reconnect behavior.
- Updated the integration version to `1.0.5`.

### Root cause

The devices could authenticate successfully but stop providing updated values because the STREAM property upload mode was not explicitly enabled. In addition, the reconnect loop marked the connection as `RECONNECTING` before calling `connect()`. The connection guard then treated that state as an already active connection and returned without starting a new BLE attempt.

### Validation

- Tested with three STREAM devices: one STREAM AC and two STREAM AC Pro units.
- The Premium setup is currently stable in the local Home Assistant installation.
- The implementation is designed for up to five devices, subject to Bluetooth adapter and radio-environment limits.
- Full test suite: `160 passed`.

The APKs and raw Home Assistant logs used for analysis are not included in this PR. The relevant reverse-engineering findings and evidence hashes are documented in the fork's `docs/reverse-engineering/STREAM_BLE.md`.

Best regards,  
Lumi (ChatGPT Luna AI)

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,8 +463,6 @@ Click on any device below to see available sensors, switches, and controls:
 
 </details>
 
-> **STREAM multi-device BLE (MrChurch fork):** This fork adds reconnect handling for multiple STREAM AC and STREAM AC Pro devices, including per-device backoff/jitter so devices do not all reconnect at the same time. STREAM telemetry is explicitly enabled after authentication and the setting is reapplied after reconnect. The reconnect state handling was also fixed so a scheduled reconnect starts a fresh BLE attempt instead of stopping in the `RECONNECTING` state. Version `1.0.5` has been tested with three STREAM devices and is designed to support up to five devices, subject to the Bluetooth adapter and radio environment. See the [STREAM BLE reverse-engineering notes](docs/reverse-engineering/STREAM_BLE.md) for the APK findings, evidence hashes, and test conclusions.
-
 <details>
 <summary><b>STREAM Microinverter</b></summary>
 
--- a/README.md
+++ b/README.md
@@ -463,7 +463,7 @@ Click on any device below to see available sensors, switches, and controls:
 
 </details>
 
-> **STREAM multi-device BLE (MrChurch fork):** This fork adds reconnect handling for multiple STREAM AC and STREAM AC Pro devices, including per-device backoff/jitter so devices do not all reconnect at the same time. STREAM telemetry is explicitly enabled after authentication and the setting is reapplied after reconnect. The reconnect state handling was also fixed so a scheduled reconnect starts a fresh BLE attempt instead of stopping in the `RECONNECTING` state. Version `1.0.5` has been tested with three STREAM devices and is designed to support up to five devices, subject to the Bluetooth adapter and radio environment.
+> **STREAM multi-device BLE (MrChurch fork):** This fork adds reconnect handling for multiple STREAM AC and STREAM AC Pro devices, including per-device backoff/jitter so devices do not all reconnect at the same time. STREAM telemetry is explicitly enabled after authentication and the setting is reapplied after reconnect. The reconnect state handling was also fixed so a scheduled reconnect starts a fresh BLE attempt instead of stopping in the `RECONNECTING` state. Version `1.0.5` has been tested with three STREAM devices and is designed to support up to five devices, subject to the Bluetooth adapter and radio environment. See the [STREAM BLE reverse-engineering notes](docs/reverse-engineering/STREAM_BLE.md) for the APK findings, evidence hashes, and test conclusions.
 
 <details>
 <summary><b>STREAM Microinverter</b></summary>
--- a/README.md
+++ b/README.md
@@ -463,6 +463,8 @@ Click on any device below to see available sensors, switches, and controls:
 
 </details>
 
+> **STREAM multi-device BLE (MrChurch fork):** This fork adds reconnect handling for multiple STREAM AC and STREAM AC Pro devices, including per-device backoff/jitter so devices do not all reconnect at the same time. STREAM telemetry is explicitly enabled after authentication and the setting is reapplied after reconnect. The reconnect state handling was also fixed so a scheduled reconnect starts a fresh BLE attempt instead of stopping in the `RECONNECTING` state. Version `1.0.5` has been tested with three STREAM devices and is designed to support up to five devices, subject to the Bluetooth adapter and radio environment.
+
 <details>
 <summary><b>STREAM Microinverter</b></summary>
 
--- a/custom_components/ef_ble/__init__.py
+++ b/custom_components/ef_ble/__init__.py
@@ -121,6 +121,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
     options = Connection.Options(
         timeout=timeout,
         bluez_start_notify=advanced.get(CONF_BLUEZ_START_NOTIFY, False),
+        max_connection_attempts=0 if device.uses_native_reconnect else 10,
+        max_reconnect_attempts=0 if device.uses_native_reconnect else 2,
+        reconnect_delay=5 if device.uses_native_reconnect else 10,
+        reconnect_delay_max=60,
+        reconnect_jitter=0.25 if device.uses_native_reconnect else 0,
     )
     issue_id = f"{entry.entry_id}_max_connection_attempts"
 
@@ -128,7 +133,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
         await (
             device.with_update_period(update_period)
             .with_logging_options(ConfLogOptions.from_config(merged_options))
-            .with_disabled_reconnect()
+            .with_disabled_reconnect(not device.uses_native_reconnect)
             .with_packet_version(packet_version.to_num())
             .with_enabled_packet_diagnostics(packet_collection_enabled)
             .with_diagnostics_on_exception(diagnostics_on_exception)
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -213,6 +213,14 @@ class Connection:
 
         timeout: int = 20
         bluez_start_notify: bool = False
+        # A value of 0 means unlimited attempts.  The integration keeps the
+        # historical defaults for regular devices and opts STREAM into a more
+        # persistent reconnect policy.
+        max_connection_attempts: int = MAX_CONNECTION_ATTEMPTS
+        max_reconnect_attempts: int = MAX_RECONNECT_ATTEMPTS
+        reconnect_delay: float = 10
+        reconnect_delay_max: float = 60
+        reconnect_jitter: float = 0
 
     _listeners = _ConnectionListeners.create()
 
@@ -344,7 +352,9 @@ class Connection:
             return
 
         max_attempts = (
-            max_attempts if max_attempts is not None else MAX_CONNECT_ATTEMPTS
+            max_attempts
+            if max_attempts is not None
+            else self._options.max_connection_attempts
         )
 
         self._connection_attempt += 1
@@ -355,7 +365,6 @@ class Connection:
                 attempts=MAX_CONNECTION_ATTEMPTS,
             )
             self._set_state(ConnectionState.ERROR_MAX_RECONNECT_ATTEMPTS_REACHED, err)
-            self._notify_disconnect(self._last_exception)
             raise err
 
         self._connected.clear()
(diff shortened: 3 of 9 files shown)


custom_components/ef_ble/eflib/connection.py
```diff
@@ -690,7 +740,10 @@ def _set_state(
         self._state_reason = reason
         self._state = state
 
-        if state.is_error:
+        if state.is_error and not (
```
GnoX, line 743: Suppressing `_notify_disconnect` while `_retry_on_disconnect` is set means an error state never reaches Home Assistant for as long as retrying continues - and with `max_reconnect_attempts=0` for STREAM, retrying never stops. So a device that is genuinely gone (powered off, out of range, taken by the app) leaves its entities looking live indefinitely, with no reload and no unavailable state, only warnings in the log.

The intent reads as "don't tear the entry down for a transient drop", which is right, but that needs a point at which the integration gives up and says so. Either keep a finite attempt ceiling for the native path, or notify once after N failures while continuing to retry in the background.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/eflib/connection.py
```diff
@@ -463,42 +476,79 @@ def _reconnect_done(task: asyncio.Task[None]):
...
+            self._options.reconnect_delay * 2 ** max(self._reconnect_attempt - 1, 0),
+            self._options.reconnect_delay_max,
+        )
+        if not self._options.reconnect_jitter:
+            return delay
+
+        # A stable offset keeps the three STREAM devices separated without making
+        # reconnect timing non-deterministic from one drop to the next.
+        device_hash = int.from_bytes(
+            hashlib.sha1(self._address.encode()).digest()[:4], "big"
```
GnoX, line 491: This is a stable per-device offset rather than jitter, which the comment is explicit about - but it does not guarantee separation. The hash is uniform, so two addresses can land arbitrarily close together and the offsets are then effectively identical; with `reconnect_jitter=0.25` the whole spread is only 25% of the delay, so at the 60 s ceiling three devices could still collide inside a couple of seconds.

A deterministic *slot* would give the guarantee the comment is after: sort the addresses and multiply the index by a fixed spacing, or serialise the connect attempts outright so only one runs at a time.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/eflib/connection.py
```diff
@@ -463,42 +476,79 @@ def _reconnect_done(task: asyncio.Task[None]):
...
+            self._set_state(ConnectionState.RECONNECTING)
+            try:
+                await self.connect()
+            except AuthErrors.BaseException as e:
+                # Authentication errors are configuration/device-state errors, not
+                # transport errors.  Retrying them forever only creates a BLE storm.
+                self._retry_on_disconnect = False
+                self._set_state(ConnectionState.ERROR_AUTH_FAILED, e)
+                return
+            except Exception as e:  # noqa: BLE001
```
GnoX, line 538: `except Exception` with a `continue` will also swallow programming errors - an `AttributeError` from a partially initialised connection, for instance - and turn them into an endless quiet retry loop, which is the hardest kind of bug to diagnose from a user's log. Catching `BleakError` plus `TimeoutError`/`OSError` would cover the transient transport cases you are after and let the rest surface.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/__init__.py
```diff
@@ -121,14 +121,19 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
     options = Connection.Options(
         timeout=timeout,
         bluez_start_notify=advanced.get(CONF_BLUEZ_START_NOTIFY, False),
+        max_connection_attempts=0 if device.uses_native_reconnect else 10,
+        max_reconnect_attempts=0 if device.uses_native_reconnect else 2,
```
GnoX, line 125: `max_reconnect_attempts=0` (unlimited) is what makes the notification suppression in `connection.py` permanent rather than temporary - see the note there. Worth reconsidering as a finite but generous ceiling, which keeps the "survive a transient drop" behaviour while still failing visibly.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/manifest.json
```diff
@@ -232,5 +232,5 @@
         "protobuf~=6.30"
     ],
 
-    "version": "1.0.3"
+    "version": "1.0.5"
```
GnoX: The version bump belongs to the maintainer's release process rather than a feature PR, and this jumps `1.0.3` to `1.0.5`, skipping a version. Dropping this file from the PR avoids a conflict with whatever the next release is numbered.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

README.md
```diff
@@ -463,6 +463,8 @@ Click on any device below to see available sensors, switches, and controls:
 
 </details>
 
+> **STREAM multi-device BLE (MrChurch fork):** This fork adds reconnect handling for multiple STREAM AC and STREAM AC Pro devices, including per-device backoff/jitter so devices do not all reconnect at the same time. STREAM telemetry is explicitly enabled after authentication and the setting is reapplied after reconnect. The reconnect state handling was also fixed so a scheduled reconnect starts a fresh BLE attempt instead of stopping in the `RECONNECTING` state. Version `1.0.5` has been tested with three STREAM devices and is designed to support up to five devices, subject to the Bluetooth adapter and radio environment. See the [STREAM BLE reverse-engineering notes](docs/reverse-engineering/STREAM_BLE.md) for the APK findings, evidence hashes, and test conclusions.
```
GnoX: This paragraph describes your fork, in upstream's README - once merged it would tell every reader of this repo about a fork that no longer differs from it. The behaviour changes themselves are worth documenting; the fork framing and the version claim are what should come out.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

GnoX (commented):
> [!NOTE]

Reviewing this as a draft, so most of the notes below are about scope rather than correctness.

First, the useful part: **your app finding independently confirms something that shipped yesterday**. A build for the STREAM AC 5000 sends exactly that write - `active_display_property_full_upload=true` on cmd_set `0xFE` / cmd_id `0x11` - and the cmd_id was the one thing in it that rested on inference rather than evidence, because that device's telemetry arrives on an unusual cmd_id. You found it in the app; two independent routes to the same conclusion is worth more than either alone. The overlap is worth knowing about before this gets merged: there is already a device-level implementation of the same write on a branch, so the question becomes whether the trigger belongs per-device or in the STREAM base.

**On the second root cause, though: that code does not run today.** `async_setup_entry` calls `.with_disabled_reconnect()` unconditionally, which sets `Connection._reconnect = False`, so `_retry_on_disconnect` is `False` for every device. `disconnected()` therefore always takes the early branch, calls `_notify_disconnect()`, and Home Assistant reloads the config entry - that reload *is* the reconnect mechanism in the shipped integration. `Connection.reconnect()` is never invoked from the integration, and no test drives it either. So the `RECONNECTING` guard could not have caused anyone's STREAM symptom; this PR is what makes that path reachable in the first place. The telemetry write is the half that explains "authenticates but values go stale".

That matters for the risk here rather than being a technicality. Today every drop notifies and the entry reload re-establishes the link; this PR moves STREAM devices onto a previously unexercised path where, with unlimited attempts, neither the notification nor the reload happens. Notes on the lines.

To be clear about the ambition: replacing entry reloads with in-place reconnect is the *right* goal - a reload tears down and rebuilds every entity, loses state and is visible in the UI. The objection is the blast radius of doing it inside `Connection` for one device family, in code that has never run in production, rather than the idea.

Three things I checked and could **not** fault: `hashlib` is already imported, `connect()` already guards with `max_attempts != 0` so a `0` really does mean unlimited, and `stream_ac_pro.Device` subclasses `stream_ac.Device` with the matching docstring, so the model check does cover both units.

The reconnect staggering also overlaps https://github.com/rabits/ha-ef-ble/issues/445, where the problem is the *initial* connect of several devices at startup rather than reconnects, so the two may want to be one mechanism. The rest is packaging: the version bump, the README paragraph and the notes file are all things a maintainer would rather own or place differently.

---
🤖 *AI-generated review (Claude Code, model Claude Opus 5 - developer on vacation until 2026-08-08). If anything here is wrong, say so - a human will follow up after Aug 8.*

docs/reverse-engineering/STREAM_BLE.md
```diff
@@ -0,0 +1,60 @@
+# STREAM BLE: Reverse-Engineering-Notizen
+
+## Zweck
+
+Diese Notiz hält die für den MrChurch-Fork relevanten Ergebnisse aus der EcoFlow-App-Analyse und den Home-Assistant-Logs fest. Der Fokus liegt ausschließlich auf STREAM AC und STREAM AC Pro sowie auf stabilen BLE-Verbindungen mit mehreren Geräten.
+
+## Lokale Beweisdateien
```
GnoX, line 7: Two things worth changing before this lands: the file is in German while the rest of the repo is English, and this section lists a local filesystem path that contains your personal account name. The APK hashes are harmless, but the path is worth removing regardless of what happens with the PR - it identifies you, and edits do not remove it from GitHub's history once pushed.

The app finding itself (the `ConfigWrite` field and the cmd_set/cmd_id pairing) is the valuable part and would survive a much shorter, English note.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/__init__.py
```diff
@@ -121,14 +121,19 @@ async def async_setup_entry(hass: HomeAssistant, entry: DeviceConfigEntry) -> bo
...
+        reconnect_jitter=0.25 if device.uses_native_reconnect else 0,
     )
     issue_id = f"{entry.entry_id}_max_connection_attempts"
 
     try:
         await (
             device.with_update_period(update_period)
             .with_logging_options(ConfLogOptions.from_config(merged_options))
-            .with_disabled_reconnect()
+            .with_disabled_reconnect(not device.uses_native_reconnect)
```
GnoX, line 136: This one line is what activates everything else in the PR. Until now this call was unconditional, so `Connection._reconnect` was `False` for every device, `_retry_on_disconnect` never became `True`, and `Connection.reconnect()` was dead code in the integration - recovery came from `_notify_disconnect()` triggering `async_schedule_reload` on the config entry.

So this is not a behaviour tweak, it is switching STREAM devices from the only recovery path that has ever run in production to one that has never run. That is worth doing eventually - entry reloads rebuild every entity and lose state - but it argues for landing the telemetry write on its own first, and treating the reconnect rework as a separate change that can be reviewed and tested as such.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/eflib/connection.py
```diff
@@ -340,11 +348,17 @@ async def connect(
         self,
         max_attempts: int | None = None,
     ):
-        if self._state.is_connecting:
+        # The native reconnect loop marks the state as RECONNECTING immediately
+        # before calling connect().  That state must enter a fresh BLE attempt;
+        # otherwise this guard returns early and reconnect() waits forever for
+        # authentication that can never happen.
+        if self._state.is_connecting and self._state is not ConnectionState.RECONNECTING:
```
GnoX, line 355: Worth knowing that this guard could not have affected users: with `with_disabled_reconnect()` unconditional in `async_setup_entry`, nothing ever set the state to `RECONNECTING`, so `connect()` was never called with it. The bug is real in the code as written, but it becomes reachable only because this PR enables the native reconnect path - it cannot be the cause of the stale-values reports.

No objection to the fix itself. It just means the "authenticates but stops providing values" symptom rests entirely on the `active_display_property_full_upload` write, which is the part I would want merged first and confirmed by the affected users.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

custom_components/ef_ble/eflib/connection.py
```diff
@@ -463,42 +476,79 @@ def _reconnect_done(task: asyncio.Task[None]):
...
-            self._set_state(
-                ConnectionState.ERROR_MAX_RECONNECT_ATTEMPTS_REACHED,
-                MaxReconnectAttemptsReached(
-                    attempts=MAX_RECONNECT_ATTEMPTS,
-                    last_error=self._last_exception,
-                ),
-            )
-            self._notify_disconnect(self._last_exception)
+        """Reconnect until the link is restored or the configured limit is reached."""
+        while self._retry_on_disconnect:
```
GnoX, line 498: Since this loop has never executed in the integration, the tests matter more than usual here - a regression test that only asserts the state transitions will pass whether or not the loop behaves on real hardware.

The cases worth covering, because each one is a way this can go wrong quietly: a drop mid-authentication, a device that never comes back (does anything user-visible ever happen?), `disconnect()` racing an in-flight reconnect, and the entry being unloaded while the loop is sleeping. That last one is where the current code sets `self._conn = None`, which other code paths then call `sendPacket` on - see issue 448.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5 - developer away until 2026-08-08). React 👎 or reply `/no-ai` to stop automated replies here.*

GnoX (commented):
Three more notes on the reconnect half, on the lines below - the short version is that the path this PR enables has never run in the shipped integration, so it deserves to be a separate change from the telemetry write.

MrChurch:
Update: The low-volume diagnostic logging has been moved to the MrChurch fork main branch for local validation and is not part of this upstream PR. The PR remains focused on the STREAM telemetry and reconnect changes. Thank you again for the detailed feedback and careful review. Best regards, Lumi (ChatGPT Luna AI)

PR #451, opened 2026-08-09
Stop devices from outbidding each other for a Bluetooth proxy and allow preferring one

### LINKED ISSUE #445: Feature request: Allow selecting a preferred Bluetooth Proxy for each EcoFlow BLE device
### Summary

Please add an option to select a preferred Home Assistant Bluetooth Proxy (Bluetooth source) for each EcoFlow BLE device.

### Why this is useful

I have multiple ESPHome Bluetooth Proxies distributed throughout my house.

I also have two EcoFlow PowerStreams that are both located next to a dedicated ESP32 Bluetooth Proxy (approximately 10 cm away).

Currently the integration uses Home Assistant's automatic Bluetooth routing. After Home Assistant or Bluetooth Proxy restarts, the PowerStreams may reconnect through a different Bluetooth Proxy instead of the dedicated nearby proxy.

### Requested feature

Add an optional configuration setting to select a preferred Bluetooth Proxy (Bluetooth source) for each EcoFlow device.

Example:

Preferred Bluetooth Proxy:
08:D1:F9:D9:6D:A6

Devices:
- PowerStream 1
- PowerStream 2

When configured, the integration should always try to establish the BLE connection through the selected Bluetooth Proxy.

If that proxy is unavailable, it could optionally fall back to Home Assistant's automatic Bluetooth proxy selection.

### Benefits

- More stable BLE connections
- Predictable reconnects after Home Assistant restarts
- Predictable reconnects after ESPHome proxy restarts
- Better performance when a dedicated proxy is installed close to the EcoFlow device
- Does not affect Bluetooth routing for other Home Assistant devices

Thank you for considering this feature.

GnoX:
This PR addresses devices connecting through a distant Bluetooth proxy while a dedicated one sits next to them. Home Assistant chooses the connection path itself - the `BLEDevice` handed to `establish_connection` is re-resolved by `habluetooth`, which ranks every candidate path by advertisement RSSI and then penalises a proxy for each connection it already has in flight, by an amount proportional to the gap between the best and second-best path. That penalty can exceed a proxy's entire signal advantage, so when two of our devices connect simultaneously the second one is scored while the first is still handshaking, and the nearby proxy loses for the whole session. Nothing here can pin a device to a proxy, but it can avoid creating that contention, and it can wait for the moment when the proxy you want is the one that would be chosen anyway.

Full list of changes:

- **A new `proxy.py` holds all of this**, so the integration entry point keeps to setting entries up: the connect gate, the preferred-proxy wait and the list of connectable proxies that the config flow offers all live there.
- **Connections are taken one at a time.** A `_connect_gate` context manager wraps the connect call in `async_setup_entry`, holding a lock shared by every entry through `hass.data`. Because a disconnect reloads the entry, reconnects pass through the same gate as initial setup.
- **Only a device that comes up alongside another waits.** Each device records when its connect finished and the next one waits out whatever remains of the window before starting, so the proxy has time to drop its in-progress count before the next device is scored. A device connecting on its own - the common case with a single device, or with devices that come up at different times - finds the window long elapsed and is not delayed at all.
- **The wait is bounded.** After 30 seconds a device gives up on the gate, logs a warning and connects anyway, so one device whose connection hangs cannot keep every other device offline.
- New **Connection delay** option under advanced connection options (seconds, 0-60, default 1), with `0` disabling the gate for that device. Its help text explains the proxy-selection effect, since the reason to raise or lower it is not otherwise guessable.
- **New optional Preferred Bluetooth proxy per device.** Since the path cannot be chosen directly, the integration instead chooses *when* to connect: it waits, for up to 20 seconds, until the selected proxy is the path Home Assistant's own ranking would pick, then connects immediately. The ranking is not reimplemented - `habluetooth`'s `score_connection_path` is called through the scanner devices Home Assistant exposes, so our idea of the best path is by construction the same as the one that will be used. A preference that cannot be honoured (proxy offline, or simply outranked) is logged and the connect proceeds normally rather than leaving the device unavailable.

Two things worth being explicit about. The preference is timing, not routing: `connect()` re-resolves the path immediately after we release it, so another integration starting a connection on that proxy in between can still change the outcome - unlikely, but not impossible. And it can only help when the preferred proxy would win once it is idle and visible; a proxy that is genuinely weaker than another will never rank first, so the wait will time out every time, which is why that case is logged plainly.

Resolves #445


ecammit (commented):
LGTM

custom_components/ef_ble/translations/uk.json
```diff
@@ -69,10 +69,14 @@
...
+              "preferred_proxy": "За можливості підключати цей пристрій через певний адаптер Bluetooth або проксі, наприклад той, що стоїть поруч із ним. Home Assistant обирає шлях самостійно, і вказати йому конкретний неможливо, тому інтеграція натомість чекає до 20 секунд, доки обраний вами проксі не стане тим, який Home Assistant обрав би й сам, і підключається саме в цю мить. Через це проксі зі слабшим сигналом, ніж в іншого, може ніколи не бути обраним. Список формується під час відкриття цієї сторінки й містить лише адаптери та проксі, які зараз у мережі, тож якщо якогось бракує, увімкніть його та відкрийте цю сторінку знову. Залишайте порожнім, якщо пристрій не підключається постійно через хибний проксі.",
```
GnoX: @megadimich Can you check these translations?

custom_components/ef_ble/translations/zh-Hans.json
```diff
@@ -69,10 +69,14 @@
...
             "data": {
               "connection_timeout": "连接超时",
+              "connection_delay": "连接延迟",
+              "preferred_proxy": "首选蓝牙代理",
               "bluez_start_notify": "使用 BlueZ start notify"
             },
             "data_description": {
               "connection_timeout": "等待设备连接和认证的秒数。",
+              "connection_delay": "当另一台 EcoFlow 设备刚刚连接完成时，本设备在连接前等待的秒数。Home Assistant 会自行选择每个连接使用哪个蓝牙代理，并会降低正在建立连接的代理的评分，因此同时启动的设备可能在整个会话期间都连接在较远的代理上。将它们错开可以让每台设备更有机会通过距离最近的代理连接。未与其他设备同时启动的设备不会被延迟。设为 0 可为该设备禁用此功能。",
+              "preferred_proxy": "尽可能通过指定的蓝牙适配器或代理连接此设备，例如就在设备旁边的那一个。Home Assistant 自行选择连接路径，无法被指定使用某一个，因此集成改为最多等待 20 秒，直到您选择的代理正好也是 Home Assistant 会选择的那一个，然后在此刻连接。因此，信号弱于其他代理的代理可能永远不会被选中。该列表在打开此页面时读取，且只包含当前在线的适配器和代理，如果缺少某一个，请将其重新上线后再次打开此页面。除非设备总是通过错误的代理连接，否则请留空。",
```
GnoX: @maybetaken Can you check these translations?

maybetaken (commented):
LGTM

custom_components/ef_ble/translations/zh-Hans.json
rabits: Resolving since we see approve from @maybetaken below.

custom_components/ef_ble/translations/uk.json
rabits: Hi @GnoX , I don't see response from @megadimich - if you think it's not needed - could you please resolve this conversation (required by automation before merge).

rabits (approved):
Looks good

custom_components/ef_ble/translations/uk.json
GnoX: As it's just a visual thing and AI seems to be getting it right, we can probably just skip confirmations and let native speaker correct it later if it turns out it's wrong.

PR #454, opened 2026-08-14
Add D365 serial prefix for the Delta 3 1500

### LINKED ISSUE #453: Device Support Request: EcoFlow DELTA 3 (D365) / Pls add support
[config_entry-ef_ble-01M00TAW6XT7TV83ZF9X7F8CJD.json](https://github.com/user-attachments/files/31083337/config_entry-ef_ble-01M00TAW6XT7TV83ZF9X7F8CJD.json)

GnoX:
Units reporting the `D365` serial prefix were listed as unsupported. `D365` is the same product the integration already handles under `D361`: EcoFlow sell it as the DELTA 3 1500 but group it under their DELTA 2 Plus product internally, so it speaks the earlier protocol generation rather than the one the DELTA 3 Max and Ultra use. Without the prefix, such a unit fell through the two-character `D3` fallback and was labelled "EcoFlow DELTA (1300)" - the 2018 model.

Changes:

- `delta2_plus.py` claims `D365` alongside `D361`.
- `_delta2_base.device` names `D365` "Delta 3 1500"; without this the unit would be adopted correctly and then displayed as "Delta 2", since the name comes from a match on the serial prefix.
- `device_mappings.py` gains the `D365` entry.

Replaying that capture through discovery selects `delta2_plus`, processes **100 of 100** packets and reports battery 99.54%, AC input 259 W (242.5 V, 1.4 A), AC output 0 W, cell temperature 32 °C, with the 1500 W AC charging ceiling applied.

Resolves #453

---
🤖 *Opened by an AI agent (Claude Code, model Claude Opus 5).*

### DIFF
--- a/custom_components/ef_ble/eflib/device_mappings.py
+++ b/custom_components/ef_ble/eflib/device_mappings.py
@@ -121,7 +121,8 @@ ECOFLOW_DEVICE_LIST = {
     # DELTA 3 FAMILY
     # =====================
     "P231":{"name": "EcoFlow DELTA 3", "packets": "v3"},
-    "D361":{"name": "EcoFlow DELTA 3 (1500)", "packets": "v3"},
+    "D361":{"name": "EcoFlow DELTA 3 (1500)", "packets": "v2"},
+    "D365":{"name": "EcoFlow DELTA 3 (1500)", "packets": "v2"},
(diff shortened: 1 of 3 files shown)


rabits (approved):
Easy!

PR #458, opened 2026-08-16
Add P322 serial prefix for the Delta 3 Classic

### LINKED ISSUE #455: [Device]: Delta 3 Classic (New Serial Number Prefix)
### Pre-submission checks

- [ ] The device is not listed in the [Supported Devices](https://github.com/rabits/ha-ef-ble#supported-devices) section of the README
- [x] I searched existing issues and there isn't already a request for this device
- [x] I've read the [Requesting Support for New Devices](https://github.com/rabits/ha-ef-ble/wiki/Requesting-Support-for-New-Devices) wiki page

### Device model

Delta 3 Classic

### Firmware version

1.0.0.0

### Serial number prefix

P322

### Bluetooth connection you tested with

Internal bluetooth adapter

### Diagnostics collection status

Attached a diagnostics dump

### Attached as diagnostics.json

[diagnostics.json](https://github.com/user-attachments/files/31091120/diagnostics.json)

### Features you care about

Same thing as the existing Delta 3 Classic install, but new serial number prefix of P322. P321 works, but the newer models of the Delta 3 Classic have a new serial number prefix. Confirmed via editing devices /eflib/delta3_classic.py that modifying the serial number to P322 worked and picked up on the device and the features were working. 

### Draft PR (optional)

_No response_

### Anything else?

_No response_

GnoX:
This PR adds the `P322` serial prefix to the Delta 3 Classic device class, so units with that prefix are detected as a Delta 3 Classic instead of falling through to the unsupported-device fallback. The app's product catalog lists `P321` and `P322` under the same `product_ps_delta_delta_3_c` key and the same product name, so it is the same product speaking the same protocol - the reporter also patched the prefix locally and confirmed the device works on hardware, and replaying their capture through `delta3_classic` parses 99 of 100 packets with 39 populated fields.

While adding the matching `ECOFLOW_DEVICE_LIST` entry, several existing Delta 3 rows turned out not to match the app catalog, so they are corrected here too. That list is only consulted for devices that have no class of their own, where it supplies the displayed name and the packet version, and for naming attached extra batteries - so these are label and lookup corrections, not protocol changes. No proto changes were needed anywhere in this PR.

Full list of changes:

- **`P322` added to `Device.SN_PREFIX` in `delta3_classic.py`.** The existing four-character prefix matching covers it, and the Delta 3 base takes its display name from the class docstring, so no name-matching logic changes.
- **`P32` added to `ECOFLOW_DEVICE_LIST`.** The catalog uses the three-character key for this model and `P321`/`P322` are the only members, so one row covers the family; the list lookup already tries progressively shorter prefixes.
- **`D3N1` and `D3M1` names were shifted by one.** The catalog and our own device classes agree that `D3N1` is the DELTA 3 Max and `D3M1` the DELTA 3 Max Plus, while the list called them Classic and Max respectively.
- **`D3E1` is an extra battery, not a device.** It was listed as "DELTA 3 Max Plus"; both the catalog and `ADDON_BATTERY_MAP` describe it as the Max Plus extra battery, and the battery map's slightly different wording is aligned with the catalog.
- **`D3U1` and `D3UP` do not exist.** The Ultra and Ultra Plus prefixes are `D751` and `D511`, which is what `delta3_ultra.py` and `delta3_ultra_plus.py` already match on, so those two rows never matched a real serial.
- **A supported device now wins over the short extra-battery prefixes when naming an attached battery.** `ADDON_BATTERY_MAP` contains two- and three-character keys broad enough to swallow whole model families - `P3` is the RIVER Pro extra battery, but it also matches every `P32x` Delta 3 Classic. Four-character battery prefixes are still checked first, so no existing battery name changes.

Resolves #455

### DIFF
--- a/custom_components/ef_ble/eflib/device_mappings.py
+++ b/custom_components/ef_ble/eflib/device_mappings.py
@@ -7,7 +7,7 @@ if TYPE_CHECKING:
 
 ADDON_BATTERY_MAP: dict[str, str] = {
     "Y712": "Delta Pro Ultra Battery",
-    "D3E1": "DELTA 3 Max Extra Battery",
+    "D3E1": "DELTA 3 Max Plus Extra Battery",
     "MR52": "DELTA Pro 3 Smart Extra Battery",
     "R361": "DELTA 2 Max Smart Extra Battery",
     "R341": "DELTA 2 Extra Battery",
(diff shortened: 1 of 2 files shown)


rabits (dismissed):
Should be ok)

rabits (approved):
Thank you!

Issue #448, opened 2026-08-06
AttributeError: 'NoneType' object has no attribute 'sendPacket' when device advertises under a rotating/alternate BLE MAC while integration still targets the original address


hardy212:
## Title
`AttributeError: 'NoneType' object has no attribute 'sendPacket'` when device advertises under a rotating/alternate BLE MAC while integration still targets the original address

## Environment
- ha-ef-ble version: v1.0.x series (HACS custom repository, `rabits/ha-ef-ble`)
- Home Assistant: 2026.7.4
- Devices: 4x EcoFlow STREAM series (3x STREAM AC Pro, 1x STREAM Ultra), connected via 2x ESPHome Bluetooth proxies
- BLE proxies: ESPHome 2026.7.4

## What happened

Intermittently, service calls that write to entities backed by this integration (e.g. `number.set_value` on a `charging_power_limit` or `feed_grid_power_limit` entity) fail with:

```
AttributeError: 'NoneType' object has no attribute 'sendPacket'
    await self._conn.sendPacket(packet)
```

This happens inside `custom_components.ef_ble.eflib.connection`. The error propagates up through Home Assistant's script/automation executor and aborts the entire running script/automation sequence at that point (not just the single failed call), even when later steps have `continue_on_error: true` set on the automation/script actions — the error appears to originate below the level `continue_on_error` can catch.

## Suspected root cause

Immediately preceding each occurrence, the log shows a `Disconnected from device` warning for the device's BLE MAC:

```
WARNING (MainThread) [custom_components.ef_ble.eflib.connection - <MAC>] Disconnected from device (_send_config_packet -> sendPacket -> sendRequest -> disconnected)
```

Cross-checking `ha_get_integration(..., diagnostics_data_path="data.bluetooth")` on the ESPHome Bluetooth proxy config entries shows that the affected device is, at the time of the error, advertising under a **different MAC address** than the one the `ef_ble` config entry was originally set up with. The advertisement payload (manufacturer ID `46517`) still decodes to the correct device serial (e.g. `BK11ZE1B2H461435` for our STREAM Ultra unit), confirming it's the same physical device — just seen under a rotating/alternate BLE address.

It looks like when the device's advertised MAC changes, the integration's internal `_conn` reference for that device entry becomes `None` (presumably during a reconnect attempt against the now-stale address), and a subsequent `sendPacket` call made before reconnection completes hits the `None` object directly instead of being queued, retried, or raising a handled/user-facing error.

## Impact

[...]

Issue #405, opened 2026-07-09
[Bug]: AttributeError: 'NoneType' object has no attribute 'call_later' during BLE reconnect


hardy212:
### Pre-submission checks

- [x] I'm running the latest release and the bug still reproduces
- [x] I searched existing issues and this isn't a duplicate

### Environment
- Integration version: v1.0.1 (also observed on v1.0.0)
- Home Assistant: HA OS on Raspberry Pi 5 (HAOS 18.x)
- Connection: ESPHome Bluetooth Proxy (2 proxies, one per DC-coupled battery pair)
- Devices affected: EcoFlow Delta Pro Ultra, 2× DC-coupled pairs (master + slave each)

### Description
When a device's BLE connection drops and reconnects (e.g. `sensor.ef_XXXXX_main_battery_level` goes `unavailable`/`unknown` and then recovers), any subsequent service call to that device (`switch.turn_on`/`turn_off`, `number.set_value`, etc.) can fail with:

```
AttributeError: 'NoneType' object has no attribute 'call_later'
```

### Traceback (representative)
```
File "/config/custom_components/ef_ble/eflib/devicebase.py", line 203, in call_later
    self._conn.call_later(delay, callback, key)
AttributeError: 'NoneType' object has no attribute 'call_later'
```

### Root cause (as far as I can tell from the outside)
`devicebase.py` line 203 calls `self._conn.call_later(...)` without checking whether `self._conn` is `None`. During a reconnect cycle there appears to be a window where the device object exists (and its entities report a fresh state) but the underlying connection object has not been fully re-established yet — any code path that hits `call_later` during that window raises.

### Steps to reproduce
1. Set up two DC-coupled EcoFlow devices behind an ESPHome Bluetooth proxy.
2. Trigger a BLE disconnect/reconnect cycle (e.g. restart the proxy, or let normal BLE instability occur — this happens periodically without any manual trigger).
3. Immediately after the affected device's sensors report a value again (reconnect complete from HA's perspective), call a service on that device (e.g. `switch.turn_on` on its `charging_task` switch, or `number.set_value` on a power-limit entity).
4. Observe the traceback above in Home Assistant's logs; the automation/script calling the service aborts (unless `continue_on_error: true` is set on the action).

### Impact
[...]

PR #459, opened 2026-08-16
Route every device packet through a connection-aware send helper

GnoX:
This PR stops writes issued during a BLE reconnect from raising `AttributeError: 'NoneType' object has no attribute ...` and takes the whole automation run down with them. `DeviceBase.disconnect()` sets `self._conn = None` as a sentinel for `observe_connection()`, so for the length of the disconnect window every send and every scheduled callback dereferences `None`. The connection layer already skipped sends on a dead link gracefully, so the crash was purely the nulled reference, not a deeper problem in the send path. No proto changes were needed.

The fix is a single `DeviceBase.send_packet()` that every packet now goes through, plus a `raise_on_failure` flag that decides what a caller wants when delivery is impossible. Background traffic keeps its existing best-effort behavior and is dropped silently; a user-initiated command raises `NotConnectedError` instead of being swallowed and reported as success, so an automation sees a clear failure rather than a write that quietly never happened.

Full list of changes:

- **`send_packet()` is now the only way a device sends anything.** It returns early when the link is gone, or raises when the caller passed `raise_on_failure=True`. There are no remaining direct `_conn.sendPacket` call sites in the library, which is what removes the crash rather than patching the paths it was reported from.
- **`raise_on_failure` is threaded through `Connection.sendPacket` / `sendRequest` / `_sendRequest`**, defaulting to `False` so every existing send behaves exactly as before. It raises in three places: no live client at send time, the link dropping mid-request, and retries being exhausted while still nominally connected. A command whose write and response both completed counts as delivered even if the link drops immediately afterwards.
- **`call_later()` is a no-op while disconnected.** This is the crash site users actually hit, since a config write commonly schedules a timer immediately after it completes. Pending callbacks are cancelled on disconnect anyway, so there is nothing to schedule.
- **All 47 send sites were ported:** 38 user-initiated commands across 18 device modules ask to fail loudly, while time sync, heartbeat polls, dormancy and wake-up traffic, protocol acks and SHP3 user-id registration stay best-effort.
- **Helpers that serve both roles take the flag.** `powerstream._send_ble_packet` and the DPU module-command helper default to raising, and their background callers (the power-pack ack and the heartbeat-info poll) opt out explicitly.
- **STREAM timer-task edits no longer keep a mod that was never delivered.** The pending modification is now removed again when the send fails, instead of being silently re-applied on the next successful edit.
- **DPU's private `_send_command_packet` was renamed to `_send_module_command`**, since sitting one underscore away from the new base method invited misreading at every call site.
- A separate commit fixes `smart_meter` and `stream_microinverter` stamping `cfg_utc_time` *after* serializing the message, so the timestamp never reached the wire.

Addresses #448, #405

### DIFF
--- a/custom_components/ef_ble/eflib/commands.py
+++ b/custom_components/ef_ble/eflib/commands.py
@@ -30,7 +30,7 @@ class TimeCommands:
         payload = utcs.SerializeToString()
         packet = Packet(0x21, 0x0B, 0x01, 0x55, payload, 0x01, 0x01, 0x13)
 
-        await self.device._conn.sendPacket(packet)
+        await self.device.send_packet(packet)
 
     async def sendRTCRespond(self):
         """Send RTC timestamp seconds and TZ as respond to device's request"""
@@ -64,7 +64,7 @@ class TimeCommands:
             0x03,
         )
 
-        await self.device._conn.sendPacket(packet)
+        await self.device.send_packet(packet)
 
     async def sendRTCCheck(self):
         """Send command to check RTC of the device"""
@@ -98,7 +98,7 @@ class TimeCommands:
             0x03,
         )
 
-        await self.device._conn.sendPacket(packet)
+        await self.device.send_packet(packet)
 
     def async_send_all(self):
         now = time.monotonic()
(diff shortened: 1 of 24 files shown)


rabits (approved):
Looks good!

PR #462, opened 2026-08-16
Route BLE notifications through a single state-dispatched handler

GnoX:
This PR replaces the per-stage notification callbacks in `Connection` with a single handler that dispatches on the current connection state, and converts the class's method names to snake_case while the call sites are being touched anyway. Every BLE notification previously landed in whichever `*Handler` method had been passed to `start_notify` for that auth stage, each wrapped in an `_auth_handler` decorator that re-checked the expected state and swapped the callback for the next one. Notifications now arrive at one `_on_notification`, which matches on the state and calls the handler for it, so the subscription is set up once and the auth stages no longer rewire the transport underneath themselves.

The renames are mechanical and behavior-preserving, and they keep the names the procedures already had: the public `sendPacket`/`sendRequest`/`replyPacket` become `send_packet`/`send_request`/`reply_packet`, and the internals gain a leading underscore. Nothing outside `eflib` calls any of them.

Full list of changes:

- **One `_on_notification` entry point replaces four stage-specific callbacks.** The `_auth_handler` decorator, which existed to assert the expected state before each callback ran, goes away: the dispatch match now decides which handler applies, so the auth state is read in exactly one place instead of being asserted independently in four, and the states and their handlers can no longer disagree about which stage is current.
- **Method names keep the app's own procedure names, in snake_case.** `initBleSessionKeyHandler` becomes `_init_ble_session_key_handler`, `getKeyInfoReq` becomes `_get_key_info_req`, `listenForDataHandler` becomes `_listen_for_data_handler`, and so on. These names come from the decompiled app, so keeping them lets the code be cross-referenced against it - only the casing changes.
- **The notify characteristic is subscribed once per connection instead of four times per handshake.** `_sendRequest` previously re-subscribed before every request that expected a reply; each of those is a BlueZ round trip that can fail synchronously - the "Remote peer disconnected" handling already in this file exists because `start_notify` does exactly that - and each is a window in which a fast device reply can land against a handler that is being replaced. This should make the handshake itself more robust, and a stage that fails can no longer leave the connection listening for a stage it has already left.
- **Adding or reordering an auth stage is now a single match arm plus its handler**, rather than a decorator, a callback swap and a state assertion that all have to agree - which is what the two PRs stacked behind this one go on to do.
- **Every inbound frame takes the same path**, so a device sending something unexpected is traceable from one entry point rather than from whichever callback happened to be subscribed at the time.
- **Public methods renamed to snake_case**, with call sites updated across `devicebase.py`, the SHP2 and SHP3 auto-reply paths, and three test modules. No public HA-facing surface changes: these are all internal to `eflib`.
- **No behavior change is intended anywhere in this PR.** The auth exchange sends the same packets in the same order; only where the responses are routed changes.

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -11,7 +11,7 @@ from collections.abc import Awaitable, Callable, Collection, Coroutine, MutableS
 from dataclasses import dataclass
 from enum import StrEnum, auto
 from functools import cached_property
-from typing import Any, Literal, Self
+from typing import Any, Literal
 
 import ecdsa
 from bleak import BleakClient
@@ -35,7 +35,6 @@ from .exceptions import (
     MaxReconnectAttemptsReached,
     NotConnectedError,
     PacketParseError,
-    PacketReceiveError,
     UnsupportedBluetoothProtocol,
 )
 from .frame_assembler import (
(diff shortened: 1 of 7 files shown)


rabits (approved):
This is a big one, will test it in main.

PR #464, opened 2026-08-16
Rewrite the auth flow as a sequential procedure and reorder Connection by lifecycle

GnoX:
This PR turns the authentication exchange into a single procedure that reads top to bottom, instead of a chain of callbacks that each had to know which stage came next. Previously every stage sent its request and returned, and the reply was picked up by a separate handler that advanced the state and started the following stage, so the order of the exchange existed only as a graph of state constants spread across the file. `_run_auth` now awaits each response where the request was sent, and the class is reordered to follow the connection lifecycle - construction, connect, auth, send, receive, teardown - rather than the order the methods happened to be written in.

Most of the diff is that reordering: of the roughly 500 changed lines, about 400 are identical lines that only moved. Reviewing with whitespace and moved-line detection on leaves around 100 lines of genuinely new code.

Full list of changes:

- **`_run_auth` drives the whole exchange.** Each stage sends and then awaits its own reply, so the sequence is visible in one function rather than inferred from which handler sets which state. A stage that raises ends the attempt in one place, with `TimeoutError` distinguished from other failures so the connection state says which happened.
- **Responses are buffered from before the request goes out.** `_expecting_response()` installs a queue and the notification dispatcher fills it while a stage is waiting; anything arriving outside a stage is device data as before. This matters more than it looks: a device can answer faster than the sending coroutine is scheduled again, and a reply that arrives with nothing waiting for it would otherwise be parsed as device data and lost, stranding the stage until it times out.
- **`_wait_notification` asserts it is called inside `_expecting_response()`**, so a future stage that forgets the buffer fails immediately instead of racing intermittently.
- **The stage helpers are named for what they do** - `_ecdh_session`, `_type_0_session`, `_type_1_session`, and the read helpers `_read_simple_response` and `_read_packet_response` - while the procedures that exist in the app keep their own names from the previous PR.
- **`Connection` is reordered by lifecycle**, which is the bulk of the diff and changes no behavior.
- No behavior change is intended: the same packets go out in the same order, the assemblers are still rebuilt for each auth routine, and the GATT validation, undecryptable-frame handling and `raise_on_failure` paths are untouched.

This is the second of three PRs splitting a larger `Connection` refactor and is based on the state-dispatched notification handler branch, so it should be read and merged after it. The decorator and typing cleanups follow separately.

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -208,7 +208,7 @@ class _ConnectionListeners(ListenerRegistry):
 
 
 class Connection:
-    """Manages client creation, authentification and sends the packets to parse back"""
+    """Manages client creation, authentication and sends the packets to parse back"""
 
     @dataclass
     class Options:
@@ -263,6 +263,8 @@ class Connection:
 
         self._tasks: set[asyncio.Task] = set()
         self._call_later_handles: dict[str, asyncio.TimerHandle] = {}
+        self._auth_task: asyncio.Task | None = None
+        self._notification_queue: asyncio.Queue[bytes] | None = None
 
         self._logger = ConnectionLogger(self)
         self._state_changed = asyncio.Event()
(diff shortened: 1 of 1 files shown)


rabits (approved):
Should be good, will check from main

PR #465, opened 2026-08-17
Record every session key in diagnostics so older captures stay readable

GnoX:
This PR records the session keys of recent connections in diagnostics, so that the packets already in the buffer stay decodable after a reconnect. A dump carries the last few hundred raw frames but only the key in use when it was taken, and a device that reconnects negotiates a fresh key: everything captured before that point decrypts to noise, which is exactly the part of a capture worth reading when the reported problem is a reconnect. The keys are recorded by the diagnostics collector, which lives on the device and therefore outlives any single connection, and they are encrypted with the same per-dump session as the existing key fields, so nothing new is exposed.

Full list of changes:

- **Every encryption change goes through one `_use_encryption()` helper**, which sets `_encryption` and emits `on_session_key_derived`. All four assignment sites use it - the type-1 derived key, the ECDH shared key, the negotiated type-7 session key, and the type-0 case, which emits nothing because there is no key.
- **The history is kept by `DeviceDiagnosticsCollector`, not by `Connection`.** A device-level reconnect builds a new `Connection` while the collector and its packet buffers stay, so keys held on the connection would be discarded exactly when the frames they decode are not. Keys and buffers now share one owner, one lifetime and one clearing point.
- **The history is capped at the ten most recent keys**, which comfortably outlives the packet buffers it exists to decode, and is cleared with them when collection is toggled.
- **The ECDH shared key is recorded too, not just the negotiated session key.** The handshake frames of a session are encrypted with the shared key, so without it the beginning of every reconnected session stays unreadable even though the rest decodes.
- **Diagnostics gain a `session_keys` list**, each entry encrypted exactly like the existing `session_key` and `iv` fields, with timestamps relative to the start of collection so they line up with the packet timestamps already in the dump.
- **Switching collection on records the key already in use**, rather than waiting for the next handshake. Collection is normally enabled after a problem appears, and the session being captured authenticated long before that, so without this the newest frames in the buffer - the ones the user just reproduced - would have no key at all.
- The history only appears in a dump when packet collection is enabled, since that is what puts frames in the buffer in the first place.

Both a BLE-level retry and a device-level reconnect now add their keys to the same history, so a dump taken after several reconnects can still be read from the first frame in the buffer. Reloading the config entry rebuilds the device, which clears the buffers and the keys together, so a dump never contains frames whose key is missing.

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -195,6 +195,7 @@ type PacketReceivedListener = Callable[[bytes], None]
 type PacketParsedListener = Callable[[Packet], None]
 type DataReceivedListener = Callable[[bytes, ConnectionState], None]
 type DataSendListener = Callable[[bytes], None]
+type SessionKeyDerivedListener = Callable[[bytes, bytes], None]
 
 
 class _ConnectionListeners(ListenerRegistry):
@@ -204,6 +205,7 @@ class _ConnectionListeners(ListenerRegistry):
     on_packet_parsed: ListenerGroup[PacketParsedListener]
     on_data_received: ListenerGroup[DataReceivedListener]
     on_data_send: ListenerGroup[DataSendListener]
+    on_session_key_derived: ListenerGroup[SessionKeyDerivedListener]
 
 
 class Connection:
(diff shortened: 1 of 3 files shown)


ecammit (commented):
LGTM

rabits (approved):
Yep, looks good!

PR #466, opened 2026-08-17
Mark auth stage entry states with a decorator and fix None-related type errors

GnoX:
This PR gives each auth stage a decorator that enters its connection state, and fixes the `None`-related type errors the previous refactors left behind. Every stage began by setting the state it represents, which is bookkeeping the reader has to check line by line to be sure a stage cannot run while the connection thinks it is somewhere else; `@_auth_stage(...)` moves that to the signature, where a stage and its state are read together. The typing half is separate from that and touches nothing at runtime: annotations that were merely absent, plus a handful of places where a value that cannot be `None` at that point was not proven so. No proto changes were needed.

Full list of changes:

- **`@_auth_stage(state)` replaces the first line of each of the four auth stages.** It is a static method on `Connection` typed with `Concatenate[Self, P]`, so a decorated stage keeps its own signature rather than collapsing to `(*args, **kwargs)`.
- **`_client` is annotated `BleakClient | None`**, which it always was, and the three `with_*` builders now return `Self` so chaining them keeps the concrete type instead of widening.
- **Three `assert` statements** where the surrounding control flow already guarantees a value is present - after a successful connect, and where the encryption exists because the handshake reached that stage. They document the invariant for a reader and for the type checker; they are not runtime checks that can be tripped by a device.
- **`MaxConnectionAttemptsReached` takes an optional `last_error`.** It was already raised without one on the path where the attempts ran out with nothing recorded, so the signature now matches how it is used.
- **The retry-exhausted branch reads more directly**: `add_error` and the re-raise both moved inside the `err is not None` guard, since neither means anything without an error to report.
- **A device advertising no name falls back to its address** in the log line, rather than logging `None`.

Nothing here changes behavior: the same states are entered in the same order, the same packets go out, and the asserts hold on every path the code can take.

This is the last of three PRs splitting a larger `Connection` refactor. It sits on the sequential auth flow branch, so it should be read and merged after it.

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -685,7 +685,7 @@ class Connection:
                 case 1:
                     await self._type_1_session()
                 case _:
-                    await self._ecdh_session()
+                    await self._init_ble_session_key()
 
             await self._auto_authentication()
             await self._wait_authenticated()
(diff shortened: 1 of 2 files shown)


rabits (approved):
Looks nice!

PR #469, opened 2026-08-18
Add ac_in_flow sensor for Delta 3 devices as a working AC-input signal

beezly:
## Summary

`plugged_in_ac` (mapped to `plug_in_info_ac_charger_flag`) never reflects whether mains power is actually present on the built-in AC inlet. That field sits alongside `plug_in_info_pv_charger_flag`, `plug_in_info_dcp_charger_flag` and `plug_in_info_dcp2_charger_flag` — it's a per-accessory "is this specific pluggable charging module connected" flag, not a general AC-input indicator. On a unit using the standard AC inlet (no swappable AC charger accessory), it simply never gets set, and the entity sits on "unknown" indefinitely even while AC power is flowing normally and other fields (`ac_input_power`, `battery_level`) keep updating fine from the same `DisplayPropertyUpload` packet.

`flow_info_ac_in` is the input-side counterpart to `flow_info_ac_out`, which is already used for the `ac_ports` (AC output on/off) sensor via the `flow_is_on` transform. This adds a new `ac_in_flow` sensor using the identical pattern for AC input, giving a reliable boolean for "is AC actually flowing in" — useful for anyone using one of these as backup/UPS power and wanting to detect a mains outage.

Verified on a Delta 3 Max by physically pulling and restoring mains power: the sensor flipped `off` within ~6s of disconnection and back `on` within ~3s of reconnection, both confirmed against `ac_input_power` dropping to/recovering from 0W in the same window.

Cross-checked field numbers and semantics against the official EcoFlow app (decompiled `com.ecoflow.energyStorageModule.proto.Pd335Sys`, same `pd335_sys` protobuf module this integration uses): `flow_info_ac_in` = field 47, `flow_info_ac_out` = field 367, `plug_in_info_ac_charger_flag` = field 202. The app's own detail-screen logic for "is AC input connected" reads `flowInfoAcIn` (OR'd with raw input power > 0 as a belt-and-braces check) — not `plugInInfoAcChargerFlag`, which the app only ever surfaces as a separate pass-through field, never as a general connectivity signal. This matches the fix here.

Scoped to `Delta3Base` only, since that's what I could test. `delta_pro_3.py`, `river3.py` and `dpu_x.py` share the identical `plugged_in_ac` mapping and would likely benefit from the same addition, but I don't have that hardware to verify against.

## Note on scope

This adds a new sensor rather than fixing `plugged_in_ac` in place. `plugged_in_ac`/`plug_in_info_ac_charger_flag` appears to be a per-accessory "is a specific swappable AC-charger module connected" flag rather than a general AC-input indicator — on hardware using the standard built-in AC inlet (no charger accessory), it never gets set and the entity sits on "unknown" indefinitely, even while AC power is flowing normally. I didn't want to repoint `plugged_in_ac` at `flow_info_ac_in` directly in case there's a device configuration where the accessory-flag field does report something meaningful and other users depend on its current behavior — that's a call for @rabits to make, not something to guess at. Happy to fold `ac_in_flow` into `plugged_in_ac` instead, or deprecate the old field, if that's preferred.

## Test plan

- [x] `python3 -m py_compile` on both changed `.py` files
- [x] `python3 -m json.tool` on the changed `en.json`
- [x] Live-verified on Delta 3 Max hardware by pulling/restoring mains power
- [x] Field semantics cross-checked against the official EcoFlow Android app's decompiled protobuf/UI logic

### DIFF
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -124,6 +124,7 @@ def shp2_channel(
 _BINARY_SENSORS: Final[dict[str, BinarySensorEntityDescription]] = {
     "error_happened": problem("error", entity_category=EntityCategory.DIAGNOSTIC),
     "plugged_in_ac": plug(),
+    "ac_in_flow": plug(),
     "fan_running": _make_desc(
         BinarySensorDeviceClass.RUNNING,
         enabled=False,
--- a/custom_components/ef_ble/eflib/devices/_delta3_base.py
+++ b/custom_components/ef_ble/eflib/devices/_delta3_base.py
@@ -81,6 +81,7 @@ class Delta3Base(DeviceBase, ProtobufProps):
     usba2_output_power = pb_field(pb.pow_get_qcusb2, out_power)
 
     plugged_in_ac = pb_field(pb.plug_in_info_ac_charger_flag)
+    ac_in_flow = pb_field(pb.flow_info_ac_in, flow_is_on)
     battery_input_power = pb_field(pb.pow_get_bms, lambda value: max(0, value))
     battery_output_power = pb_field(pb.pow_get_bms, lambda value: -min(0, value))
 
--- a/custom_components/ef_ble/translations/en.json
+++ b/custom_components/ef_ble/translations/en.json
@@ -340,6 +340,9 @@
       "plugged_in_ac": {
         "name": "AC Plugged In"
       },
+      "ac_in_flow": {
+        "name": "AC Input Flowing"
+      },
       "fan_running": {
         "name": "Fan Running"
       },


GnoX (commented):
> [!NOTE]
> Quick first pass from an AI agent working in the maintainer's repo, so this doesn't have to wait for his next free evening: it has the full history, his notes and the protocol reverse-engineering material to work from, and strict instructions on what it may change, publish and claim. GnoX reviews these and corrects anything that turns out wrong.

Thanks for the write-up - the mains-pull timing test is the sort of evidence that makes a field mapping reviewable, and the field numbers check out: `flow_info_ac_in = 47` sits beside `flow_info_ac_out = 367`, and `ac_ports` already uses the same `flow_is_on` pattern.

I ran your two claims against the diagnostics captures already collected for other Delta 3 models. The short version: **the sensor is a good idea, but `flow_is_on` is the wrong transform for the input field**, and the premise about `plug_in_info_ac_charger_flag` holds on some models and not others.

**87 paired samples of `flow_info_ac_in` and `pow_get_ac_in`, across three models:**

| model | `flow_info_ac_in` | AC input measured | `flow_is_on` says |
|---|---|---|---|
| Delta 3 Plus | `15` (`0b01111`) | 133-159 W | on - correct |
| Delta 3 Classic | `12` (`0b01100`) | 186-192 W | **off - wrong** |
| Delta 3 Ultra Plus | `4` (`0b00100`) | 0 W | off - correct |

`flow_is_on` keeps only the low two bits, so `12` reads as off while that unit was charging from the wall at ~190 W. On those samples the sensor would report no AC input while AC input is flowing. **Bit 3 (`x & 0b1000`) agrees with the measured power on all 87 samples** and the three distinct values above. Worth saying plainly: that is fitted to the captures we have, not confirmed against the app's own check, so treat it as the better hypothesis rather than the answer.

**On `plug_in_info_ac_charger_flag`:** it is not universally missing. On the Plus and the Classic it is present, `True`, and repeated throughout the capture - 33 messages in 35 seconds on the Classic, while AC was flowing. It is **absent** from the Ultra Plus capture, so there is at least one model that behaves as you describe, and your Max may be another.

**And there is a mechanism that produces a permanently "unknown" entity even when the flag is present.** A property notifies listeners only when its value *changes* - `Field._set_value` returns early if the new value equals the current one. An entity snapshots the current value when it is constructed but registers its update callback later, in `async_added_to_hass`. For a field that stays `True` for as long as mains is connected, a notification landing in that window is lost and nothing repeats it, so the entity keeps the "unknown" it started with. Fields that keep changing, like `ac_input_power`, repair themselves on the next update - which is the asymmetry you described.

So there may be two separate things here: a missing field on some models (real, worth a sensor) and an entity that never picks up a constant value (a plumbing bug worth fixing centrally).

**What would settle your side of it:** a diagnostics dump from the Max with packet collection enabled while mains is connected - Diagnostics -> Enable packet collection in the device options, a few minutes, then the JSON from the device page. That shows whether the flag is absent on a Max, and lets the `flow_info_ac_in` value be compared against `ac_input_power` on your own hardware, which is the check that matters for the transform.

---

**Correction to the above.** I have since decrypted two more Delta 3 captures - a Max Plus and a second Ultra Plus - and they refute the bit-3 suggestion I made. `flow_info_ac_in = 12` appears on a Delta 3 Classic while it draws 187 W from the wall, and on a Delta 3 **Max Plus** while it draws **0 W**. Same value, opposite states, so no bitmask over that field can be correct on both; my "agrees with all 87 samples" was an artifact of not having the Max Plus data in front of me.

What the wider set does show is that `plug_in_info_ac_charger_flag` tracks AC input far better than `flow_info_ac_in` does. Across eight captures spanning Classic, Plus, Max Plus, Ultra Plus and Delta Pro 3, it is `True` in every capture carrying 105-678 W of AC input and `False` in the one carrying 0 W. One honest limit on that: `pow_get_ac_in = 0` distinguishes not-charging from charging, not unplugged from plugged, so the single `False` capture cannot be told apart from a unit that is plugged in and simply idle at 99% - the flag's negative case is not actually verified by these captures. What supports reading it as presence rather than charging is the Classic, which reports `True` at 100% SoC with `cms_chg_dsg_state = 0` while drawing 187 W.

| model | `flow_info_ac_in` | `plug_in_info_ac_charger_flag` | AC input |
|---|---|---|---|
| Delta 3 Classic | `12` | `True` | 186-192 W |
| Delta 3 **Max Plus** | `12` | `False` | **0 W** |
| Delta 3 Plus (two captures) | `15` | `True` | 105-159 W |
| Delta 3 Ultra Plus | `14` | `True` | 364-382 W |
| Delta 3 Ultra Plus (solar) | `4` | **absent** | 0 W |
| Delta Pro 3 (two captures) | `2` | `True` | 366-678 W |

So the shape of the fix changes. The flag is the reliable indicator, and the reason it reads "unknown" on your unit is most likely the notify-on-change plumbing described above rather than the field being wrong. The one genuine gap is the Ultra Plus solar capture, where the flag is not sent at all - that is where a fallback signal would help, and it is a narrower problem than this PR currently addresses. On this evidence I would not ship `ac_in_flow` derived from `flow_info_ac_in` with any mask.

A dump from your Max would still be worth having: if the flag is present and `False` there while mains is connected, that is a different fault again and worth chasing on its own.

---
🤖 *AI-generated review comment (Claude Code, model Claude Opus 5). React 👎 or reply `/no-ai` to stop automated replies here.*

beezly:
[config_entry-ef_ble-01KZZR1X6XSBGTX99VFGTR0DPE.json](https://github.com/user-attachments/files/31195249/config_entry-ef_ble-01KZZR1X6XSBGTX99VFGTR0DPE.json)

Find the diagnostics attached.

GnoX:
Thanks for the dump. It answers both open questions, and on the first one it moves things in your favour.

**Your premise is confirmed, and more firmly than the evidence I had before.** Across the 95 `0xFE/0x15` frames in your capture, `plug_in_info_ac_charger_flag` is **absent from every single one**, while `pow_get_ac_in` reads **292-323 W** throughout. That matters because the only other capture with the flag missing was an Ultra Plus sitting at 0 W, where "plugged in but not charging" could not be ruled out. Yours cannot be explained that way: mains is plainly charging and the flag never arrives. So there is a real missing-signal gap on at least one model, and a fallback sensor has something genuine to fill.

It also explains your symptom without needing the notification race I mentioned. On your unit the entity is not losing an update, the field simply never arrives. That race is still a real bug for fields that *are* sent and never change, but it is not what you are hitting.

**And a correction to my own review: `flow_is_on` is right on your hardware.** Your `flow_info_ac_in` is `14`, which the transform reads as on while you were charging. Scored against every model and state I can pair with a measured input, it is correct in six of seven combinations:

| `flow_info_ac_in` | `flow_is_on` | measured AC input | model |
|---|---|---|---|
| `15` | on | 105-159 W | Delta 3 Plus |
| `14` | on | 292-323 W | **Delta 3 Max (yours)** |
| `14` | on | 364-382 W | Delta 3 Ultra Plus |
| `2` | on | 366-678 W | Delta Pro 3 |
| `4` | off | 0 W | Ultra Plus (solar) |
| `12` | off | 0 W | Delta 3 Max Plus |
| `12` | **off** | **186-192 W** | Delta 3 Classic |

So my flat "wrong transform" was too strong, and I should not have put it that way. The single failure is the last row, and it is not fixable by choosing different bits: `12` means charging on a Classic and idle on a Max Plus, so the same value carries both meanings and no mask can separate them. The defect is a false negative confined to the Classic, not a flaw in the idea.

That is what remains to decide, and it is the maintainer's call rather than mine: as written the sensor is correct for your model but would report no AC input on a Classic drawing ~190 W. Reporting *unknown* for the ambiguous value instead of `off` would keep your case working without asserting something false on a Classic, which seems the honest shape if it ships for all Delta 3 devices.

One thing to keep off the table: `plug_in_info_ac_in_flag` (field 61) reads `1` on your Max while charging, but `0` on a Delta Pro 3 charging at 366-678 W. Whatever it tracks, it is not mains presence, so it should not be mapped until someone establishes what it actually means.

---
🤖 *AI-generated comment (Claude Code, model Claude Opus 5).*

GnoX:
Can I ask what is the point of this sensor? Does it do something else than `ac_input_power > 0` that you can use in automations directly? `plug_in_info_ac_charger_flag` is different in that it is True while it is plugged in even if it not charging, but I see that this one is false while this one is discharging. 

I get that it is not available on newer firmwares and Max models and we can remove it, but I don't really see a point in adding binary sensor that just basically does the same thing as input power, just with less information and its actual bitfield not completely understood.

PR #475, opened 2026-08-23
Add per-device options, apply the app's charging minimums with per-device unlock options, add a Delta 2 charge pause switch

### LINKED ISSUE #421: [Sensor]: Delta 2 Max - PauseFlag
### Pre-submission checks

### Device model

Delta 2 Max

### Sensor / control name

chgPauseFlag

The sensor/switch pauses the AC charging and is useful in HA automations when you want to charge via AC e.g. in case of PV surplus.
I use it in my Automations for that already but currently using the HTTP API, hopefully it is in the BLE available too (for information in original Ecoflow_Cloud integration it was not exposed too)?

I can set it via HTTP / API by:
```
elif cmd == "SET_CHG_PAUSE_FLAG":
            if not (value == "0" or value == "1"):
                return {"unsupported value (0/1)": value}
            params = {"chgPauseFlag": int(value), "chgWatts": 200 }
            if DEBUG: LOGGER.debug(params)
            body = {
                "id": 123, 
                "version": "1.0", 
                "sn": sn, 
                "moduleType": 5, 
                "operateType": "acChgCfg",
                "params": params
            }
```

Ecoflow Documentation:
https://developer-eu.ecoflow.com/us/document/PP?id=2058828363647848450 

For information: it is in the documentation mentioned for the Delta 2 too, but on Delta 2 it was not working for me. Now I no longer have a Delta 2. So I think it should be Delta 2 Max specific.

### Integration version

1.0.2

### Diagnostics

if diagnostics needed, I need to learn to dump first ...

### Anything else?

_No response_

GnoX:
This PR gives the AC charging speed and DC charging current controls the same minimums the EcoFlow app enforces, and adds a per-device advanced option for each so a user who wants to go lower still can. Until now those controls had no lower bound at all on most devices, which is how the 1 W AC charging trick came about: it stops the device drawing from the grid without unplugging it, and it is what solar-surplus automations have been using. The app's floors are per model rather than per family - it dispatches its AC charging range on the serial prefix, giving the Delta 3 and Delta 3 Plus 100 W where the rest of the pd335 family gets 200 W - so each device class declares its own value instead of inheriting one from a base that cannot know which model it ended up on. Because a floor would take the 1 W trick away from Delta 2 owners, this also adds the AC charging pause switch that does the job properly, reading `chg_pause_flag` (Delta 2) / `cfg_pause_flag` (Delta 2 Max) and toggling only the trailing pause byte of the AC-charge-set command the speed control already uses.

Full list of changes:

- **`DeviceOption` and `DeviceBase.ADVANCED_OPTIONS` are the whole mechanism.** A device class declares the options it offers, the integration passes the configured values in with `with_advanced_options()`, and the device reads one back with `advanced_option()`. Values live under a `device_options` key in the entry options and are picked up on an options-flow save without a restart.
- **`option_field()` turns a boolean option into a field.** Option field reads as either value depending on the option, and declaring it also registers the option in the owner's `ADVANCED_OPTIONS`, so there is no second declaration to keep in sync.
- **Two options rather than one**, `unlock_ac_charging_minimum` and `unlock_dc_charging_minimum`, so a device only advertises the option for a control it actually has - Delta 2 and Delta Pro 3 offer the AC one alone. Both are unset by default, which means the app's minimum applies.
- **AC minimums applied:** 200 W across the Delta 2 series and most of the pd335 family, 100 W on the Delta 3 and Delta 3 Plus, 400 W on the Delta Pro 3, 50 W on the River 3 family. The models that subclass a 100 W class without sharing its floor (Delta 3 Max, Max Plus, Ultra) re-declare 200 W, since our class hierarchy nests the larger models under the smaller ones.
- **AC charging pause switch for Delta 2 and Delta 2 Max**, exposed as an `AC Charging` switch: on resumes charging, off pauses it, with the watt field of the command left unchanged so the configured speed survives a pause.
- **Limit fields renamed to a `max`/`min` suffix** for consistency with the fields that already read that way. This is a rename of internal fields only - none of them back an entity - and it updates the deprecated number platform's string props and the existing device tests that reference them.

This is a behavior change for anyone already writing a sub-minimum value: an automation that sets 1 W on a Delta 3 or River 3 will now land on the app minimum until the matching option is ticked. Delta 2 owners get the pause switch as the better replacement for that trick.

Resolves #421


rabits (approved):
Nice change.

PR #479, opened 2026-08-24
Return no controls for devices without UpdatableProps instead of raising

javasoup:
### The bug

`get_controls()` goes through `get_updatable_prop_device()`, which raises
`TypeError` when a device does not implement `UpdatableProps`:

```python
def get_updatable_prop_device(device: DeviceBase):
    if not isinstance(device, UpdatableProps):
        raise TypeError("Device has to be subclass of UpdatableProps")
    return device

def get_controls[E: controls.ControlType](device, control_type) -> list[E]:
    return get_updatable_prop_device(device).get_controls(control_type)
```

`UnsupportedDevice(DeviceBase)` has no props mixin, so for any device that falls
back to it **every control platform aborts during setup**:

```
ERROR ... Error while setting up ef_ble platform for button: Device has to be subclass of UpdatableProps
TypeError: Device has to be subclass of UpdatableProps
ERROR ... Error while setting up ef_ble platform for switch: ...
ERROR ... Error while setting up ef_ble platform for number: ...
ERROR ... Error while setting up ef_ble platform for select: ...
ERROR ... Error while setting up ef_ble platform for climate: ...
```

All five fail the same way, so none of that device's entities are created and
the log gets five tracebacks per reload.

I hit this on a Power Kits Power Hub while it was being detected as
`UnsupportedDevice`. Sensors worked; every control entity was missing.

### The change

A device with no updatable props simply has **no controls** — that is a normal
state, not an error. Return an empty list so the platforms add nothing and setup
completes:

```python
def get_controls[E: controls.ControlType](device, control_type) -> list[E]:
    if not isinstance(device, UpdatableProps):
        return []
    return device.get_controls(control_type)
```

`get_updatable_prop_device()` is left raising, for callers that genuinely
require props.

### Tests

`tests/eflib/test_get_controls.py` covers all five control types plus the
unchanged raising behaviour.

Verified it actually catches the bug:

```
against current code:  5 failed, 1 passed
with this change:      6 passed
```

No regressions — `tests/eflib` gives **156 passed** both with and without the
change.

<sub>Note: 4 pre-existing failures in `tests/eflib/test_dpu.py` appear in my
environment identically with and without this change (protobuf version
difference against the pinned `uv` venv), so they are unrelated.</sub>


tests/eflib/test_get_controls.py
GnoX, line 1: Not sure if any of these unit tests is useful to have on this layer, the impact of all changes here is only for unsupported devices that are just for debugging so it does not matter if control registration crashes, I would just not include the test at all here.

custom_components/ef_ble/eflib/__init__.py
```diff
@@ -72,7 +72,18 @@ def get_updatable_prop_device(device: DeviceBase):
...
     device: DeviceBase, control_type: type[E]
 ) -> list[E]:
-    return get_updatable_prop_device(device).get_controls(control_type)
+    """
+    Return the device's controls of the given type.
+
+    A device that does not implement `UpdatableProps` simply has no controls -
+    `UnsupportedDevice` is the obvious case. Raising here would abort setup of
+    *every* control platform (button, climate, number, select, switch) for that
+    device, so return an empty list instead and let the platforms add nothing.
```
GnoX, line 81: Uh, sure, why not explain a very simple 2 line condition using a 4 line paragraph that includes info from one layer above and one below that this function should not be aware of. 

This is just AI slop comment that is not going to help anyone, remove it.

custom_components/ef_ble/eflib/__init__.py
```diff
@@ -72,7 +72,18 @@ def get_updatable_prop_device(device: DeviceBase):
 def get_controls[E: controls.ControlType](
     device: DeviceBase, control_type: type[E]
 ) -> list[E]:
-    return get_updatable_prop_device(device).get_controls(control_type)
+    """
+    Return the device's controls of the given type.
```
GnoX, line 76: Wrong docstring style, should not end with period as per contribution rules
```suggestion
    Return the device's controls of the given type
```

GnoX (approved):

rabits (approved):
Looks good!

PR #487, opened 2026-08-30
Fix auth handshake race that can corrupt the session on connect

ecammit:
## Summary

Fixes a race in the auth handshake introduced by #464's rewrite to a sequential procedure. A notification that arrives before the handshake has a session key to decrypt with - whether during the scheduling gap before the auth task runs, before `_start_notify()` itself returns (e.g. a ghost BlueZ subscription left over from a dropped connection), or just unsolicited telemetry - was being handed to `_parse_enc_packets` as if it were ordinary device data. That either hit an assertion (no encryption at all yet) or, worse, decrypted with a stale encryption object left over from a previous session, producing garbage that looked like a corrupt frame and eventually tripped the undecryptable-frame limit, forcing a reconnect that hit the same failure again - a loop only a physical BLE dongle replug would break. Reproduced consistently on a Delta Pro Ultra.

**Fix (per @GnoX's suggestion):** guard the actual precondition where it's violated, rather than chasing every timing gap a notification could arrive early in. `_listen_for_data_handler` now drops any notification if `self._connection_state.received_session_key` is false, with a debug log, instead of handing it to `_parse_enc_packets`. The auth stage actually waiting for a reply still reads it via the existing per-stage `_expecting_response()` buffer, unaffected. Net diff against `main` is 15 lines in `connection.py`.

## Test plan
- [x] `pytest tests -q` - 160 passed
- [x] New tests in `tests/eflib/test_connection_auth_notification_queue.py` cover: the per-stage buffer's queue/drain behavior, dropping vs. parsing based on session-key state, and an end-to-end `connect()` test where a stray notification fires during `_start_notify()` and the handshake still proceeds without error
- [ ] Confirmed against real hardware (Delta Pro Ultra / SHP2) that this resolves the periodic dongle-replug-required auth failures


custom_components/ef_ble/eflib/connection.py
```diff
@@ -955,23 +968,22 @@ async def _on_notification(
...
-            self._notification_queue = None
-            # Unread notifications are still device data, and the assembler needs
-            # every byte in order.
-            while not queue.empty():
-                await self._listen_for_data_handler(queue.get_nowait())
+        if self._notification_queue is None:
+            raise RuntimeError(
+                "auth stages must run within _run_auth()'s notification buffer"
+            )
+        yield
```
GnoX: **🔴 Blocking · correctness: dropping the buffer instead of draining it loses device data and can desync the frame assembler.**

The context manager being replaced ended with `while not queue.empty(): await self._listen_for_data_handler(queue.get_nowait())`, under the comment "Unread notifications are still device data, and the assembler needs every byte in order". Nothing takes that over: the new `finally` only assigns `None`, and nothing reads between stages, so whatever accumulated is discarded.

Verified both ways: on `7cde8e59` a notification left in the buffer reaches `_listen_for_data_handler` exactly once, on this head the handler is never called. That handler feeds a stateful `FrameAssembler`, so dropping the leading chunks of a multi-notification frame leaves the continuations to be parsed mid-frame, and the failures land in `add_error`.

Fix: drain into `_listen_for_data_handler` in the `finally` before clearing, as the old code did.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -479,6 +479,11 @@ async def connect(
 
         if self._auth_task is not None and not self._auth_task.done():
             self._auth_task.cancel()
+        # Open the buffer here, synchronously, rather than inside `_run_auth()` - that
+        # coroutine is only scheduled, not awaited inline, and a fast-answering device
+        # can notify before it actually gets to run for the first time. Anything that
+        # arrives in that gap must still land in the queue, not `_listen_for_data_handler`.
+        self._notification_queue = asyncio.Queue()
```
GnoX: **🔴 Blocking · correctness: a cancelled previous auth task nulls the buffer this line just created.**

`Task.cancel()` two lines up only schedules the `CancelledError`. The old `_run_auth()` resumes on a later loop iteration, unwinds through the new inner `finally`, and runs `self._notification_queue = None` unconditionally, wiping the queue installed here for the new handshake. `disconnect()` into a quick reconnect reaches the same place through `_cancel_tasks()`.

Reproduced: with an auth task parked in a stage, `cancel()` followed immediately by a fresh `asyncio.Queue()` assignment, no await in between exactly as here, leaves `_notification_queue is None` after one loop turn. The next stage then hits the new guard, raises `RuntimeError("auth stages must run within...")`, and `_run_auth` turns that into `_auth_failed`, so the reconnect fails outright.

Fix: keep the queue this run created in a local and clear the attribute only while it is still that object.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -955,23 +968,22 @@ async def _on_notification(
...
         """
-        Buffer notifications for an auth stage instead of handing them to `_handle_data`
-
-        The buffer has to exist before the request goes out: a device can answer faster
-        than this coroutine is scheduled again after the write, and a reply that arrives
-        with nothing expecting it is parsed as device data and lost, which strands the
-        stage until it times out.
+        Assert the connection-wide notification buffer is open for this auth stage
+
+        The buffer spans the *whole* handshake - opened in `connect()` before
```
GnoX: **🟠 Should fix · correctness: widening the buffer trades the assertion for a silently misparsed reply.**

With the queue live from `connect()`, anything the device emits between subscription and the pub-key write is already waiting when `_ecdh_key_exchange()` calls `_read_simple_response()`, and the buffer cannot tell a stage reply from spontaneous telemetry. Confirmed by putting one unsolicited payload in the queue before the request: it comes straight back as the ECDH response, so `_get_ecdh_type_size(data[2])` runs on device data.

In fairness the base is not correct here either, since the same notification used to reach `_listen_for_data_handler` with no encryption set, which is the reported assertion. The point is that this window is re-symptomed rather than closed, and a wrong `_dev_pub_key` is harder to recognise than an assert.

Fix: drain what predates the request before awaiting the reply, or have each stage skip notifications that do not match its expected response shape.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -479,6 +479,11 @@ async def connect(
 
         if self._auth_task is not None and not self._auth_task.done():
             self._auth_task.cancel()
+        # Open the buffer here, synchronously, rather than inside `_run_auth()` - that
```
GnoX: **🟠 Should fix · correctness: the race is narrowed, not closed, because the queue is created after `_start_notify`.**

The subscription is awaited at line 468, before this block. BlueZ can deliver notifications while the `start_notify` reply is still in flight, so those callbacks run inside that await, find `_notification_queue is None`, and fall through to `_listen_for_data_handler` with no encryption set: the reported `assert self._encryption is not None`.

Reproduced on this head by delivering one notification from inside `_start_notify` during `connect()`. It reaches the data handler, because the queue does not exist yet. The comment's own reasoning applies one step earlier than the code, since the gap that matters opens at subscription, not at task creation.

Fix: assign the queue immediately before `await self._start_notify(...)`, and clear it on that call's failure path so a failed connect leaves nothing behind.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

tests/eflib/test_connection_auth_notification_queue.py
```diff
@@ -0,0 +1,155 @@
+import asyncio
```
GnoX: **🟠 Should fix · tests: this file pins the implementation without covering the bug, and I would drop it entirely.**

Every test here replaces methods of the class under test (`_listen_for_data_handler`, `_init_ble_session_key`, `_auto_authentication`, `_wait_authenticated`, `_run_auth`) and then asserts that an attribute holds the value the line above set, or that a guard clause added in the same diff raises. Four of them assert the same `finally` from different angles.

The cost is real: all seven pass on this head while probe tests confirm four live defects in the same diff, including the deleted drain, so the file reads as coverage where there is none. It also freezes design decisions that are the least settled part of the change, which is what makes the next refactor expensive. Every other file under `tests/eflib/` tests device message parsing rather than `Connection` internals.

Fix: delete the file. If a connection test is wanted, one that drives `connect()` against a fake client which pushes a notification during `start_notify` and asserts the link still authenticates would test behavior, and would fail today.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -679,13 +684,21 @@ async def _run_auth(self) -> None:
...
+            try:
+                match self._encrypt_type:
+                    case 0:
+                        await self._type_0_session()
+                    case 1:
+                        await self._type_1_session()
+                    case _:
+                        await self._init_ble_session_key()
+            finally:
+                # From here on, the auth reply (and everything after) is confirmed by
```
GnoX: **🟡 Nit · readability: one decision is explained three times, and the test docstrings date themselves.**

The comment at line 482, this comment and the rewritten `_expecting_response` docstring all say the same thing: the buffer is connection-wide now, opened early because `_run_auth` is only scheduled. That is 19 of the 35 lines added to this file against 16 lines of code, and the surrounding methods are far sparser.

The docstring is the natural single home, except it now describes `_run_auth`'s lifecycle rather than what the context manager does, so the explanation sits under the wrong function.

Several test docstrings are notes to this review rather than to a future reader: "Regression test: an earlier version of this fix...", "see the regression this guards against", "the reported bug". A year from now nobody can find an earlier version of a fix.

Fix: keep the rationale once, in the docstring, cut the two inline comments to a line each or drop them, and state constraints rather than history in any docstring that survives.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent, every finding re-verified with probe tests run against this head and its base), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

GnoX (changes requested):

GnoX:
I can't reproduce it easily myself, but the diagnosis seems to be correct. However, I think there is a simpler fix for this problem - the reason an early notification is fatal is not the routing, it is that `_parse_enc_packets` builds the frame assembler eagerly and `_create_frame_assembler` needs an encryption that does not exist yet. So instead of widening the notification buffer so nothing can reach that path early, I would check the precondition where it lives: if there is no session key yet, drop the bytes with a debug log. A few lines, and it holds whenever the notification arrives, including before `_start_notify` returns, which the current buffer still does not cover.

I can take over if you want.

ecammit:
> I can't reproduce it easily myself, but the diagnosis seems to be correct. However, I think there is a simpler fix for this problem - the reason an early notification is fatal is not the routing, it is that `_parse_enc_packets` builds the frame assembler eagerly and `_create_frame_assembler` needs an encryption that does not exist yet. So instead of widening the notification buffer so nothing can reach that path early, I would check the precondition where it lives: if there is no session key yet, drop the bytes with a debug log. A few lines, and it holds whenever the notification arrives, including before `_start_notify` returns, which the current buffer still does not cover.
> 
> I take over if you want.

@GnoX I'll give it a look first, since I can somewhat reproduce. Basically I just keep randomly getting one DPU or SHP2 device that disconnects (for whatever reason) and it can't seem to recover. It'll hit the 10 consecutive errors limit and never resolve unless I do a full reboot or disconnect the dongle. Sometimes the full system reboot doesn't even work.

ecammit:
Went with your suggestion - reverted the connection-wide buffer entirely back to the original per-stage `_expecting_response()` (with its drain-on-exit intact) and added just the precondition guard: `_listen_for_data_handler` now drops any notification if `self._connection_state.received_session_key` is false, with a debug log, instead of handing it to `_parse_enc_packets`. That's the actual failure point either way - no encryption yet (assert), or a stale encryption object left over from a previous session decrypting garbage after a reconnect (which looks closer to what my log shows).

This also holds before `_start_notify()` returns, since it's a state check rather than a timing fix, so it covers the gap the buffer-widening approach didn't.

Net diff against `7cde8e5` is now 15 lines in `connection.py`. Force-pushed to this branch and to `main` on my fork. Should be much easier to review than what was there before - let me know if you still want to take it from here, otherwise happy to keep iterating.

(written by Claude)

ecammit:
I'm going to let this simmer overnight to see if the connection issues I've been having resolve and will take out of draft and re-request review if it's good.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -1003,6 +1003,21 @@ async def _read_packet_response(self) -> list[Packet]:
             return packets
 
     async def _listen_for_data_handler(self, data: bytes):
+        if not self._connection_state.received_session_key:
```
GnoX: **🟠 Should fix · design: the guard drops the bytes before anything records them, so a dump loses the evidence.**

The predicate looks right to me. I replayed our recorded type 1 and type 7 sessions against this head: both still decode past auth, a reconnect holding the previous link's key drops the stale bytes, and a pre-key notification no longer raises. `received_session_key` also resets with the link, so it needs no second flag to stay honest.

The placement costs something. `self._listeners.on_data_received(...)` is the first line of `_parse_enc_packets`, so returning above it means dropped notifications never reach diagnostics: I patched that listener and it is never called. When someone reports a reconnect loop, the dump then shows a gap where the bytes that caused it should be. Moving the same check into `_parse_enc_packets`, immediately after that listener line, keeps them in the dump and also covers `_read_packet_response`, which reaches the same assembler by another path.

Fix: keep the predicate, move the check one level down.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent; every claim re-checked by running this head and its base against recorded device captures), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

custom_components/ef_ble/eflib/connection.py
```diff
@@ -1003,6 +1003,21 @@ async def _read_packet_response(self) -> list[Packet]:
...
+            # Ordinary device data can't exist yet - there's no session key to decrypt
```
GnoX: **🟡 Nit · readability: eight lines of comment argue the case where two would state the rule.**

The block explains the reasoning, both alternative outcomes and the design of the buffer path, and the `log_filtered` call underneath then says it a third time in its message. What a reader needs is the constraint: the assembler for encryption types 1 and 7 is built from the session key, and this link has not derived one yet.

Fix: cut to the constraint and let the log line carry the rest.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (fork agent; every claim re-checked by running this head and its base against recorded device captures), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

GnoX (changes requested):

tests/eflib/test_connection_auth_notification_queue.py
GnoX: This is still not resolved, this really limits internal implementation and just makes refactoring harder than it needs to be, especially with agentic development. Just remove it.

ecammit:
Addressed all three outstanding review threads in 7f7e751:

- Moved the guard from \`_listen_for_data_handler\` into \`_parse_enc_packets\`, right after the \`on_data_received\` listener call - dropped notifications now stay in the diagnostics dump, and this also covers \`_read_packet_response()\`'s path to the frame assembler.
- Trimmed the guard's comment down to the actual constraint (session key needed to build the frame assembler for types 1/7), letting the log line carry the rest.
- Deleted \`tests/eflib/test_connection_auth_notification_queue.py\`.

Net diff against \`main\` is now 11 lines in \`connection.py\` plus the deleted test file. CI is green (\`Run python tests\`, \`validate-code-style\` passed; \`validate\`/\`validate-hacs\` still running as of this comment).

PR #488, opened 2026-08-30
Make the Wave 2 climate entity's temperature unit follow the device's display unit

ecammit:
## Summary

The Wave 2's climate entity temperature reads correctly only when the physical unit's LCD is set to Celsius - switching the device to Fahrenheit makes the reported temperature "way off."

`wave2.py` already detects the device's active display unit (`temp_unit`, derived from `temp_sys`) and already uses it for the target-temperature step/min/max via `unit=dynamic(temp_unit)` on both `@_climate.target_temp(...)` and `@controls.temperature(...)`. But the `controls.climate(...)` call itself never passed `temperature_unit=`, so the climate entity's own declared unit stayed hardcoded at the framework default of Celsius regardless of what the device is actually reporting.

A captured Fahrenheit heartbeat in the existing test suite confirms the device already reports `ambient_temperature`/`target_temperature` pre-converted into whichever unit its LCD is showing (`63.62` decodes as a plausible room temperature in °F, not °C) - so no additional F↔C conversion math is needed, only telling the entity which unit that already-correct value is in.

`ClimateBuilder.temperature_unit()` already supports a dynamic field/`DynamicValue` and already converts it via `unit_to_hassunit()` - this capability just wasn't wired up for any device yet.

## Fix

Add `temperature_unit=dynamic(temp_unit)` to the existing `controls.climate(...)` call in `wave2.py`, reusing the same `dynamic(...)` helper already used elsewhere in the file.

## Test plan
- [x] `pytest tests -q` - 156 passed
- [x] New test asserting the climate control's `temperature_unit` is a `DynamicValue` wrapping `temp_unit` (not the fixed default) and resolves to the correct unit against both a Celsius and a Fahrenheit captured heartbeat
- [ ] Confirmed against real Wave 2 hardware with the LCD set to Fahrenheit

### DIFF
--- a/custom_components/ef_ble/eflib/devices/wave2.py
+++ b/custom_components/ef_ble/eflib/devices/wave2.py
@@ -178,6 +178,7 @@ class Device(DeviceBase, RawDataProps):
             "high": FanGear.HIGH,
         },
         current_temperature_field=ambient_temperature,
+        temperature_unit=dynamic(temp_unit),
     )
 
     @_climate.power(power)
--- a/tests/eflib/test_wave2.py
+++ b/tests/eflib/test_wave2.py
@@ -11,6 +11,7 @@ from custom_components.ef_ble.eflib.devices.wave2 import (
     SubMode,
     WaterLevel,
 )
+from custom_components.ef_ble.eflib.entity.base import DynamicValue
 
 PACKETS = {
     "fan_celsius_target_30": "aa026c00bc2de6b30200012d42214250e4e5f8e45a3990a7e6ece6e7e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e4e7e7e7e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e66d8b9fa7e6e6e4e6e6e6e7e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e6e671a6",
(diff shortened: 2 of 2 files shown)


custom_components/ef_ble/eflib/devices/wave2.py
```diff
@@ -178,6 +178,7 @@ def _climate_main_mode(self) -> MainMode | None:
             "high": FanGear.HIGH,
         },
         current_temperature_field=ambient_temperature,
+        temperature_unit=dynamic(temp_unit),
```
GnoX, line 181: **🔴 Blocking · correctness: this line is a no-op, and the new test passes without it.**

`climate.target_temp(..., unit=...)` already assigns `self.temperature_unit = unit` (`eflib/entity/controls.py:415`), and `wave2.py:241` already passes `unit=dynamic(temp_unit)` to `@_climate.target_temp(...)`, which binds onto this same `_climate` object as the class body executes. The wiring downstream is complete too: `ClimateBuilder.temperature_unit()` resolves a `DynamicValue` through `_get_field` and stores `temperature_unit_prop`, which drives the `_attr_temperature_unit` update callback. Built from the control on `main` at this PR's base (`7cde8e59`), with the `wave2.py` hunk reverted:

```
temperature_unit_prop = 'temp_unit'
min_temp_prop = 'target_temperature_min'   # 60 in F, 16 in C
current_temperature_prop = 'ambient_temperature'
```

The new test then passes on that same reverted tree (`2 passed`, whole file 37 passed), because it asserts a class attribute the framework already sets rather than any behavior this diff changes.

The decode reasoning in the description is right, incidentally: the C fixtures decode `ambient_temperature` around 15.4 °C and the F ones around 62.7 °F, the same room in both units. That is why `main` already labels the entity correctly, since commit 32273691 introduced `temp_unit`, the 60/86 limits and the `wave_temperature()` sensors.

Fix: drop the change, since it cannot be what makes Fahrenheit read correctly or incorrectly. If a Wave 2 on `main` still reports a temperature that is way off with the LCD in Fahrenheit, the cause is somewhere else, so it would help to say which entity is wrong, what it shows against the LCD, and to attach a diagnostics dump taken in that state.

---
🤖 AI-generated review comment, [Claude Code](https://claude.com/claude-code) `/code-review` (single agent, branch checked out in a scratch worktree, test and entity description re-run against reverted code), model Claude Opus 5 (1M context, `claude-opus-5[1m]`). Reviewed by @GnoX before posting.

ecammit: Totally correct, sorry for the wasted tokens reviewing this. It was actually the Ecoflow Cloud integration that had broken temp, I just pointed Claude at the wrong one and didn't verify the work as I should have since I was distracted by other branches I'm working on. Closing this AI slop.

PR #489, opened 2026-08-30
Read each auth reply from one notification inbox instead of the next frame that arrives

GnoX:
This PR gives `Connection` a single notification inbox and makes each auth stage read from it until its own reply arrives, instead of treating the next frame that shows up as the answer. That fixes the silent handshake failures reported on a River 3 Plus and a Delta 2 Max: some units send each handshake reply about seven times, copies of the previous step land after the next request has gone out, and the session-key stage accepted one as key material. It then raised `KeyInfoReqFailed`, which is invisible at default verbosity, so the connection just stopped. 1.0.4 was immune by accident, since it called `stop_notify` between stages and the duplicates arrived while nothing was subscribed; making the subscription permanent exposed them.

Full list of changes:

- **One `asyncio.Queue` holds notifications, and who reads it depends only on where the connection is.** `_on_notification` records the frame and enqueues it, deciding nothing. The running stage reads during the handshake, one data pump afterwards, retired with a sentinel rather than a cancellation, which would land inside device parsing and half-apply a packet.
- **A stage accepts only its own reply**, by type and by the size its curve requires, and hands anything else to the data path while it keeps waiting. Each stage parses with its own assembler, so a foreign fragment cannot be spliced onto what the next stage reads.
- **A frame the link cannot decode yet is dropped rather than fed to the assembler.** The state that opens the decode gate is set after the session key is installed, not before it is derived.
- **Teardown is bounded end to end**, so an entry unloading while a parser holds a packet no longer blocks Home Assistant, and a stage handing work to the data path is shielded from its own cancellation for a bounded moment only.
- **A data-path failure is reported through one policy**, from the pump or from a stage, and one arriving after the link already ended with a reason no longer replaces that reason.

Exercised on a River 3, over an internal BlueZ adapter and over an ESPHome proxy, with faults injected at runtime rather than by editing the branch: a device's own public-key reply replayed seven times into the session-key window (the failure this PR is about), captured frames delivered before the session key exists, a notification arriving from inside `start_notify`, a subscribe call raising the way BlueZ raises it, undecodable frames driven past the session-loss limit, the data path held eight seconds mid-packet and then reloaded, a packet held twelve seconds inside device parsing while the entry is unloaded, the first auth request swallowed, two reloads a second apart, adoption with a wrong key, the proxy rebooted under a live link and its route taken away and given back. 23 runs across the two transports, all passing, with no leaked futures, no cancellation landing mid-packet and no real failure reason replaced by a generic one; the held-packet unload completes in 1.1 s and 3.8 s, against 14.7 s before this change.

### DIFF
--- a/custom_components/ef_ble/eflib/connection.py
+++ b/custom_components/ef_ble/eflib/connection.py
@@ -57,6 +57,8 @@ MAX_CONNECTION_ATTEMPTS = 10
 # (notably through an ESPHome proxy). Left unbounded it stalls `async_unload_entry`
 # long enough for HA to mark the entry `FAILED_UNLOAD`, so cap every disconnect.
 DISCONNECT_TIMEOUT = 5.0
+# Waited before DISCONNECT_TIMEOUT rather than inside it, so a teardown costs the sum
+PUMP_STOP_TIMEOUT = 1.0
 
 
 _BT_PROTOCOL_UUIDS = {
@@ -246,7 +248,6 @@ class Connection:
         self._encrypt_type = encrypt_type
         self._encryption: EncryptionStrategy | None = None
         self._initial_session_key: bytes = b""
-        self._simple_assembler = SimplePacketAssembler()
         self._frame_assembler: FrameAssembler | None = None
         self._options = Connection.Options()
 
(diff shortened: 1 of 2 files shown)


rabits (approved):
Looks nice! Thank you.

PR #496, opened 2026-09-06
Gate Wave 2 controls on power state and withhold select options the device rejects

GnoX:
This PR stops the Wave 2 offering controls the device is not currently able to act on. Several writable controls could be changed while the unit was off or in standby, where the device ignores the write and the entity silently reverts a moment later; Drain Mode's "Drain Free" is only accepted while cooling; and Power Mode's "Off" is turned back into "Standby" by the firmware whenever the unit runs on mains. All of it is decidable from `power_src`, a field the device class already parsed but never used.

Selects needed one new capability for this: an option that is valid in one device state and rejected in another can neither be excluded permanently, which loses a working feature, nor handled with `availability`, which hides the options that still work.

Full list of changes:

- **`option_set` replaces the `exclude` keyword on `controls.select`,** carrying both permanent exclusions and ones conditional on a device field: `options=controls.option_set(PowerMode).exclude(PowerMode.INIT).exclude_if(plugged_in_ac, PowerMode.OFF)`. `exclude=` had a single call site, so nothing else moved.
- **`exclude_if` accepts a boolean field or a `dynamic(field, transform)`,** so Drain Mode can express a test over `main_mode` inline rather than through a computed field. `EcoflowSelect` recomputes its options from one update callback per condition, the mechanism `availability` already uses; a select with no conditions keeps a fixed list and behaves as before.
- **`power_src` is decoded into AC Plugged In, Solar Connected and Battery Connected** through the existing `prop_has_bit_on` transform. The old comment had mains and battery the wrong way round, which never showed because the field was commented out.
- **Ambient Light, Automatic Drain, Drain Mode, Fan Speed, Main Mode, Sub Mode and Target Temperature now follow `power`** and report unavailable outside the running state instead of accepting a write and reverting.

Both new behaviours rest on readings from a single unit. The `power_src` bits come from a unit reporting 1 on mains alone, 16 on battery alone, 17 on mains and battery and 18 on battery and solar, which fixes bits 0, 1 and 4 but leaves bit 2 unexercised; the Power Mode gating withholds Off only while mains is present, the case confirmed to be demoted, so units on solar alone still offer it.

Supersedes #490, whose reporting and hardware readings this is built on.

### DIFF
--- a/custom_components/ef_ble/binary_sensor.py
+++ b/custom_components/ef_ble/binary_sensor.py
@@ -124,6 +124,8 @@ def shp2_channel(
 _BINARY_SENSORS: Final[dict[str, BinarySensorEntityDescription]] = {
     "error_happened": problem("error", entity_category=EntityCategory.DIAGNOSTIC),
     "plugged_in_ac": plug(),
+    "battery_connected": connectivity(),
+    "solar_connected": connectivity(),
     "fan_running": _make_desc(
         BinarySensorDeviceClass.RUNNING,
         enabled=False,
--- a/custom_components/ef_ble/eflib/devices/wave2.py
+++ b/custom_components/ef_ble/eflib/devices/wave2.py
@@ -8,7 +8,7 @@ from ..props import computed_field
 from ..props.enums import IntFieldValue
 from ..props.raw_data_field import dataclass_attr_mapper, raw_field
 from ..props.raw_data_props import RawDataProps
-from ..props.transforms import pround
+from ..props.transforms import prop_has_bit_on, pround
 
 pb = dataclass_attr_mapper(KT210SAC)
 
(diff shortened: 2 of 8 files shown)


GnoX:
@ecammit Cleaned up your #490, can you check if I didn't break something? [v0.45.0-e.1](https://github.com/GnoX/ha-ef-ble/releases/tag/v0.45.0-e.1)

ecammit:
Yes, everything is working as expected. Tested:

- "Off" only appears when on battery-only (and works... unit disconnects from bluetooth/wifi and requires physical power button press to reconnect)
- AC connected, battery connected, and solar connected sensors all work.
- Drain free only appears in cool mode
- Target Temp, ambient light, automatic drain, drain mode, fan speed, main mode, and sub mode are all unavailable when unit is in standby.

rabits (approved):
Looks perfect!

Issue #494, opened 2026-09-05
Replace insecure `ecdsa` dependency


JamieMagee:
`ecdsa` is marked insecure for [CVE-2024-23342](https://github.com/tlsfuzzer/python-ecdsa/security/advisories/GHSA-wj6h-64fc-37mp). The `ecdsa` maintainers confirm that key generation and ECDH are affected and do not plan to fix the vulnerability.

Could we replace `ecdsa` with a side-channel-resistant implementation that supports the SECP160r1 ECDH used for Bluetooth key exchange?

GnoX:
I don't think this is applicable for what we use `ecdsa` for - we just use it for ECDH, never for signatures which is what I understand that CVE is all about. Also we only do a single ECDH derivation at the start of connection - the key itself is ephemeral and used exactly once. But supposing that it was possible to apply Minerva itself (which seems to to be nearly impossible on HA machine from what I understand), you would only be able to use that key to read traffic for that exact session so even if it could be applied, its attack sufrace would be very limited.

Or am I missing something important?

**edit**: I did try to search for alternernatives, but I can't actually find anything that we could use - pyca/cryptography/PyCryptodome only expose 192 bit curves at the shortest (Ecoflow devices do need 160 bits, that is out of our hands), some I see are outdated and probably also contain this vulnerability or don't provide wheels for python 3.14. I wouldn't mind replacing it, but I am not able to find anything suitable - do you have any suggestions?

JamieMagee:
I've opened https://github.com/rabits/ha-ef-ble/pull/498 using OpenSSL 3's EVP API through [`ctypes`](https://docs.python.org/3/library/ctypes.html). It supports SECP160r1 without changing the handshake. Local tests pass, but I don't have a device for hardware testing.

GnoX:
You seem to really care about this to the point you let an agent implement a 1000 line PR, but **can you at least provide a good reason of why we should replace it**? I would like to explain it better to our consumers than "You now need a system dependency because of vulnerability that would take 30 000 years to exploit".

[...]

PR #498, opened 2026-09-07
Replace ecdsa with OpenSSL for BLE key exchange

JamieMagee:
Use OpenSSL 3's EVP API through `ctypes` for the existing SECP160r1 handshake, replacing the `ecdsa` dependency that is marked as insecure.

Refs [#494](https://github.com/rabits/ha-ef-ble/issues/494).

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -7,7 +7,10 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
-    name: Run python tests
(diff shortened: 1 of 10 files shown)


rabits (changes requested):
Ok, the change is very big and introducing another dependency with no proper explanation. Let's use #494 to discuss and understand it better, before do sudden movements that can harm security rather then improve it.

rabits:
Closing for now, please reopen when the questions in #494 will be resolved.

PR #508, opened 2026-09-12
Put the Wave 3 into standby from the climate entity and report its temperatures in Celsius

GnoX:
This PR makes the Wave 3's climate entity turn the unit off into standby instead of powering it down, wakes it again from the same entity, and reports its temperature sensors in the unit the device actually sends. `cfg_power_off` cuts the Bluetooth radio along with the unit, so a Wave 3 switched off from Home Assistant could only be switched back on at the unit itself; the app's own home-screen toggle uses `cfg_sys_pause` for off and `cfg_power_on` for on, which keeps the link up. Powering down is still available, as a separate button that is disabled by default.

Full list of changes:

- **The climate off state sends the unit to standby, and on wakes it.** This is what the app does from the same control, and it leaves the device reachable, so the entity that turned it off can turn it back on.
- **Power off moved to its own button, disabled by default.** Pressing it drops the Bluetooth link with the unit, which is a deliberate action rather than a side effect of an off state.
- **A mode written to a sleeping unit carries the wake with it**, so calling the mode setter directly wakes the unit too rather than writing into a device that is still in standby.
- **The submode select is withheld while the unit is in standby**, since a unit in standby reports no operating mode for a submode to belong to.
- **Wave 3 temperature sensors report Celsius.** The device sends Celsius regardless of the unit its own display is set to, so the Fahrenheit setting was relabelling the value without converting it: an ambient temperature of 22.5 °C was shown as 22.5 °F. Home Assistant converts for display, so a user whose system is set to Fahrenheit now sees the right number.
- **Waking uses `cfg_power_on`.** An earlier attempt used a `cfg_sys_resume` field that looked like the obvious counterpart to `cfg_sys_pause`; regenerating the schema from the current EcoFlow app showed the device has no such field, which is why waking that way did nothing.

Resolves https://github.com/rabits/ha-ef-ble/discussions/470

### DIFF
--- a/custom_components/ef_ble/eflib/devices/wave3.py
+++ b/custom_components/ef_ble/eflib/devices/wave3.py
@@ -132,7 +132,12 @@ class Device(DeviceBase, ProtobufProps):
 
     @computed_field
     def is_submode_available(self) -> bool:
-        return self.operating_mode in (OperatingMode.COOLING, OperatingMode.HEATING)
+        # A unit in standby reports no mode, so the submode it belongs to is not
+        # selectable either
+        return self.power is True and self.operating_mode in (
+            OperatingMode.COOLING,
+            OperatingMode.HEATING,
+        )
 
     @classmethod
     def check(cls, sn):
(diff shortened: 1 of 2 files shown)


rabits (approved):
Yep, approved!

